PR #186, opened 2017-09-18
[WIP] Fix #164 kmeans _ptr methods

### LINKED ISSUE #164: kmeans needs _ptr and return results only on GPU as pointers as well
E.g., predictions or clusters should be saved on GPU and default for _ptr methods is to only have those things on the GPU.

mdymczyk:
* Added/modified most of the Python code for KMeans

Still needs work:

* Uploading data (single matrix)
* Freeing pointers need checking - might work the GLM stuff
* C++/CUDA backend for KMeans needs tweaking to handle GPU pointers - will base my code on GLM but not sure how much I can reuse since they are using C++ Matrix classes there


mdymczyk:
Probably will have to be redone after all the changes.

PR #223, opened 2017-10-01
Build non-nccl version on jenkins

pseudotensor:
Add build and s3 push for non-nccl version of xgboost during jenkins pipeline.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -245,21 +245,21 @@ pipeline {
                     def _buildVersion = version[1]
                     version = null // This is necessary, else version:Tuple will be serialized
 
-                    #if (isRelease()) {
-                    #    def artifact = h2o4gpu-${versionTag}-py36-none-any.whl
-                    #    def localArtifact = src/interface_py/dist2/${artifact}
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -183,6 +183,88 @@ pipeline {
...
+
+                script {
+                    CONTAINER_NAME = "h2o4gpu${SAFE_CHANGE_ID}-${env.BUILD_ID}"
+                    // Get source code
+                    withCredentials([[$class: 'AmazonWebServicesCredentialsBinding', credentialsId: "awsArtifactsUploader"]]) {
+                        try {
+                            sh """
+                                    nvidia-docker build -t opsh2oai/h2o4gpu-build -f Dockerfile-build .
+                                    nvidia-docker run --init --rm --name ${CONTAINER_NAME} -d -t -u `id -u`:`id -g` -v /home/0xdiag/h2o4gpu/data:/data -v /home/0xdiag/h2o4gpu/open_data:/open_data -w `pwd` -v `pwd`:`pwd`:rw --entrypoint=bash opsh2oai/h2o4gpu-build
+                                    nvidia-docker exec ${CONTAINER_NAME} rm -rf data
```
mmalohlava: @abal5 could it be the `exec`- part replaced by a script? Each stage doing the same, seems like it could be easily squashed to a single command.

PR #236, opened 2017-10-13
Fixes #177 Improved matrix multiplication for skinny matrices

### LINKED ISSUE #177: KMeans extremely slow for low k
```
from h2o4gpu import KMeans as h2oKMeans
from sklearn.cluster import KMeans as skKMeans
from sklearn.datasets.samples_generator import make_blobs
from sklearn.metrics.cluster import v_measure_score

n_samples = 1000000
centers = 16
# Generate random 'center' centroids and 'n_samples' samples around them
X, true_labels = make_blobs(n_samples=n_samples, centers=centers,
                             cluster_std=1., random_state=42)

kmeans_h2o = h2oKMeans(n_gpus=1, n_clusters=centers, max_iter=5, tol=1e-8)
kmeans_h2o.fit(X)
```

Fit takes around `Wall time: 3.84 s`. If we change centers to `17` we get `Wall time: 543 ms`.

The difference clearly starts to show at around `75000` rows.

mdymczyk:
Cublas matrix multiplications that were used to calculate distances between data points and centroids are not optimized for tall and skinny matrices, which means that if either number of rows was very high compared to a number of dimensions or the number of clusters was very small the computation was very slow. After some manual benchmarking it seems that cublas is good if k is bigger than 16 and number of dimensions is bigger than 64.

A specialized matrix multiplication was introduced for that case, using shared memory. This probably still can be tuned to use shared memory and number of threads better but even with this simple improvement, we went from being x1.5-x2 slower than SKLearn to being x2.5-x4 faster.

### DIFF
--- a/src/gpu/kmeans/kmeans_impl.h
+++ b/src/gpu/kmeans/kmeans_impl.h
@@ -165,7 +165,7 @@ int kmeans(
     }
 
     // Scale the centroids on host
-    if(n_gpu > 1) {
+    if (n_gpu > 1) {
       //Average the centroids from each device
       for (int p = 0; p < k; p++) h_counts[p] = 0.0;
       for (int q = 0; q < n_gpu; q++) {
@@ -179,7 +179,7 @@ int kmeans(
       for (int p = 0; p < k; p++) {
         for (int r = 0; r < d; r++) {
           if (h_counts_tmp[p] != 0) {
-            h_centroids[p*d + r] = 0.0;
+            h_centroids[p * d + r] = 0.0;
           }
         }
       }
@@ -191,7 +191,7 @@ int kmeans(
         for (int p = 0; p < k; p++) {
           for (int r = 0; r < d; r++) {
             if (h_counts_tmp[p] != 0) {
-              h_centroids[p*d + r] += h_centroids_tmp[p*d + r];
+              h_centroids[p * d + r] += h_centroids_tmp[p * d + r];
             }
           }
         }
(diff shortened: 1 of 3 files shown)


pseudotensor:
Looks good.  Is it clear from profiling that actually gemm is always issue?  What was going on with odd or even cases changing performance alot, ruining 4x and only giving 1x sklearn.  Kmcuda says they get more like 30x sklearn (sklearn is single threaded afterall).  Be good to benchmakr against kmcuda and/or intell dall at least in python level in test script

mdymczyk:
WIP - need to fix compilation problem.

mdymczyk:
@pseudotensor yes, `gemm` is the major offender. Depending on the data size we are getting a x2.5 - x40 times speedup. I think the biggest issue now is that this improvement is very simple, we probably can improve how much shared memory we are using and how many threads - think that's the problem for odd/even sizes.

But I'd prefer to merge this (after it compiles and tests pass) and look into a different improvement - in principle you shouldn't need to calculate all the distances from each point to each centroid at each iteration, we should be able to just compute the norm vector for the data matrix (once since it doesnt change), the norm vector for centroid matrix (at each iteration) and then for each data point compute the distance only between that point and the centroid for `min(centroid_i.norm - row.norm)`. With this we don't need the `pairwise_distances` matrix (for fitting and predictions, only for transform) and we would cut the number of distance calculations by `k`. Oh and we wouldn't be even using the method that is the current bottleneck.

Just need to figure out a way to do it all efficiently.

pseudotensor:
Sounds cool

PR #237, opened 2017-10-13
[WIP] KMeans|| implementation

### LINKED ISSUE #229: Implement kmeans++ (kmeans||)
Currently only random initialization is supported which does result in subpar optimization, kmeans++ (kmeans||) should help with that.

mdymczyk:
Fixes #229

Randomized centroid initialization for K-Means can often lead to very poor performance. For example running it with `k=4, n=100 000, d=7`:

```
from h2o4gpu import KMeans
from sklearn.metrics.cluster import v_measure_score
from sklearn.datasets.samples_generator import make_blobs

centers = 4
X, true_labels = make_blobs(n_samples=100000, centers=centers, cluster_std=1., random_state=42, n_features=7)
kmeans_h2o = KMeans(n_gpus=1, n_clusters=centers, random_state=42)
kmeans_h2o.fit(X)
v_measure_score(kmeans_h2o.labels_, true_labels)
```

Will yield only a 50% accuracy.

The idea behind improved initialization methods such as KMeans++ is to choose initial centroids which are as far apart from each other as possible using probabilities.

KMeans|| build upon KMeans++ ideas but instead of choosing only 1 centroid in each iteration (which can be slow for larger ks) it chooses several potential centroids at each step.

The final step might require running a weighted KMeans++ should we find more than k potential centroids but that KMeans++ should run on a very small set (only on potential centroids) so should be fast.


mdymczyk:
@pseudotensor 

> Is the init responsible for even odd performance issue?

No, I'm still looking into that. I want to implement the improvement mentioned by NVidia guys (no need to compute all distances at each iteration) - it should fix that issue.

> How do you test init, just by comparing resulting accuracy, or actual unit test for init vs (say) init from sklearn

For starters I was thinking about adding python tests comparing all 3: accuracy for h2o "random", accuracy for h2o "kmeans||" and accuracy for sklearn "kmeans++". Run them X times and see what is the result - in theory both kmeans|| and kmeans++ should result in better accuracy than random.

I also want to add C++ unit tests - the idea behing kmeans++ and kmeans|| is to choose points that are as far apart as possible, this should be possible to unit test by computing the total distance between them and checking if there's a better configuraiton.

src/gpu/kmeans/kmeans_h2o4gpu.cu
```diff
@@ -172,6 +178,456 @@ void random_centroids(int verbose, int seed, const char ord,
...
+ * @param data_dots
+ * @param centroids
+ * @param weights
+ * @param pairwise_distances
+ * @param labels
+ */
+template<typename T>
+void count_pts_per_centroid(
+    int verbose,
+    int num_gpu, int rows_per_gpu, int cols,
```
RAMitchell: Prefer passing a reference to a vector of vectors to using pointer pointers here.

src/gpu/kmeans/kmeans_h2o4gpu.cu
```diff
@@ -252,6 +708,7 @@ int kmeans_fit(int verbose, int seed, int gpu_idtry, int n_gputry,
   thrust::device_vector<int> *labels[n_gpu];
   thrust::device_vector<T> *d_centroids[n_gpu];
```
RAMitchell: Use vector of vectors here as noted before.

src/gpu/kmeans/kmeans_h2o4gpu.cu
```diff
@@ -172,6 +178,456 @@ void random_centroids(int verbose, int seed, const char ord,
...
+ * @param data
+ * @param weights
+ * @param centroids
+ */
+template<typename T>
+void add_centroid(int idx, int cols,
+                  thrust::host_device<T> &data,
+                  thrust::host_device<T> &weights,
+                  thrust::host_device<T> &centroids) {
+  for (int i = 0; i < cols; i++) {
```
RAMitchell: These kind of manipulations should be avoided on large vector types as they trigger resizing reallocating and copying. If the vectors are small it could be ok.

src/gpu/kmeans/kmeans_h2o4gpu.cu
```diff
@@ -172,6 +178,456 @@ void random_centroids(int verbose, int seed, const char ord,
...
+
+template<typename T>
+int kmeans_fit(int verbose, int seed, int gpu_idtry, int n_gputry,
+               size_t rows, size_t cols, const char ord,
+               int k, int max_iterations, int init_from_data,
+               int init_data, T threshold,
+               const T *srcdata, void **pred_centroids, void **pred_labels);
+
+template<typename T>
+int pick_point_idx_weighted(
```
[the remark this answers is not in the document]
mdymczyk: host_device is me coding on a plane after taking Vicodin :-) changed, to `host_vector`

src/common/matrix/utils.h
```diff
@@ -0,0 +1,19 @@
+/*!
+ * Copyright 2017 H2O.ai, Inc.
+ * License   Apache License Version 2.0 (see LICENSE for details)
+ */
+#pragma once
+
+template<typename T>
+void self_dot(std::vector<T> array_in, int n, int dim,
+              std::vector<T>& dots);
+
```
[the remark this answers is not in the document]
mdymczyk: hm, would moving it to `kmeans_utils` would be better? Since it will be used by both the GPU and CPU implementations.

src/gpu/kmeans/kmeans_h2o4gpu.cu
mdymczyk: yes, will benchmark it - the idea is that this whole method should only be called as the last step of kmeans||, where the dataset is small (only potential centroids).

tests/gpu/kmeans/test_kmeans_h2o4gpu.cu
```diff
@@ -0,0 +1,41 @@
...
+#include <thrust/host_vector.h>
+
+TEST(KMeans, CountsPerCentroids) {
+  // GIVEN
+  int k = 2;
+  int d = 2;
+  int n = 4;
+
+  // Setup data
+  thrust::host_vector<float> h_data(n*d);
```
RAMitchell: Use C++ initializer lists to succinctly init a vector.

tests/gpu/kmeans/test_kmeans_h2o4gpu.cu
```diff
@@ -0,0 +1,41 @@
...
+  h_data[0] = 0.0f; h_data[1] = 0.0f; // [0,0]
+  h_data[2] = 0.0f; h_data[3] = 0.0f; // [0,0]
+  h_data[4] = 1.0f; h_data[5] = 1.0f; // [1,1]
+  h_data[6] = 1.0f; h_data[7] = 1.0f; // [1,1]
+  thrust::device_vector<float> *d_data = new thrust::device_vector<float>();
```
RAMitchell: It is best practice to never manually allocate memory (new/malloc) in c++11. This device vector can just be a local variable. This means it is on the stack and gets freed at the end of scope. There are two memory leaks here as new is called without calling delete.

PR #241, opened 2017-10-19
Add TruncatedSVD

### LINKED ISSUE #49: Standard and Truncated SVD
Standard SVD on GPU is in place without API, but holding until can see how to do truncated SVD.  Also, seems sklearn truncated SVD uses lots of memory despite being truncated -- can one optimize more for memory vs. time?

navdeep-G:
Resolves #49 
Utilizes `cusolverDnSsyevd` to compute eigenvalues and eigenvectors, which can be used to get singular values, the U matrix, etc.

Some plots to display reconstruction error and speed benchmarks (conducted on random data):
tsvd_k100

tsvd_k49_n50

tsvd_bench


src/gpu/tsvd/tsvd.cu
```diff
@@ -12,6 +12,14 @@
...
 {
 
+/**
+ * Division utility to get explained variance ratio
+ *
+ * @param XVar
+ * @param XVarSum
+ * @param ExplainedVarRatio
+ * @param context
+ */
```
RAMitchell: Possible divide by zero in this function?

mdymczyk:
Will try this out myself later in the day but looks really good 👍  We should strive to make the whole repo like this.

src/gpu/tsvd/tsvd.cu
navdeep-G: ah, good point. will fix.

navdeep-G:
Thank you! @mdymczyk @RAMitchell

navdeep-G:
@RAMitchell Right now the python tests run through pytest, but no consistent way of writing them. And unifying matrix utilities would be great. I would love to have a single data representation instead of several across algos.

mdymczyk:
@navdeep-G @RAMitchell one more question - from what I see this is a single GPU implementation, right? Is it possible to distribute SVM across GPUs (either for speed or when matrices are too big)?

RAMitchell:
Possibly could do something with cublasXT.  Have never used it before.

pseudotensor:
It sucks major.  Very slow

PR #249, opened 2017-10-25
Switch from virtualenv to pyenv to fix problem with fpectl needed due…

pseudotensor:


mdymczyk:
Does this mean our bundle won't work with virtualenv? (not talking about docker here)

pseudotensor:
No

PR #250, opened 2017-10-25
fix/ src/interface_py/Makefile: fix link errors

memory-yancy:
Fix soft links error in xgboost lib, or it will a dead links, and then failed.

### DIFF
--- a/src/interface_py/Makefile
+++ b/src/interface_py/Makefile
@@ -38,14 +38,14 @@ all: prep
 	#cat ../../requirements_runtime.txt > requirements.txt
 
 	# for incorporating xgboost and py3nvml builds
-	ln -sf ../../xgboost/python-package/xgboost .
+	ln -sf ../../xgboost/xgboost .
 	ln -sf ../../py3nvml/py3nvml .
 
 	python setup.py sdist bdist_wheel
 
 	# update build with xgboost shared library
 	mkdir -p build/lib/xgboost/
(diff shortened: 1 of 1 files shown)


mdymczyk:
@pseudotensor does this look ok? We're still having problems with Jenkins builds for outside PRs so I'll have to run and check it manually.

pseudotensor:
No, it looks wrong in general

memory-yancy:
connect the first change and the second change to look, xgboost install in `xgboost` directory, and the `../../../../../xgboost/lib/libxgboost.so` doesn't exist.

pseudotensor:
Since things work previously for a long time, the question is what is different about your system vs. our systems.  If we made your change, ours would stop working.  E.g., xgboost is definitely in python-package as originally.

memory-yancy:
OK,  I know reason ... See https://github.com/h2oai/h2o4gpu/issues/252. Since `xgboost/python-package` will be got when run `make fullinstall`, so `make fullinstall` will solve it and others things,  but I run `make runtime` directly according to README.md file. When I find a error, I debug it and solve it . So I know reason now and I will close it .

pseudotensor:
Hi @memory-yancy Thanks.  I guess the README only says to do make runtime for the docker.  If you are saying the docker build steps are wrong, then @navdeep-G can fix that as he added it recently.

PR #251, opened 2017-10-25
Add performance test for tsvd

navdeep-G:
* Add performance unit test for tsvd, which will compare h2o4gpu and sklearns randomized algorithm in terms of timing

### DIFF
--- a/tests_open/test_tsvd_bench.py
+++ b/tests_open/test_tsvd_bench.py
@@ -1,0 +1,81 @@
+import numpy as np
+import time
+import sys
+import logging
+from sklearn.decomposition import TruncatedSVD as sklearnsvd
+from h2o4gpu.solvers import TruncatedSVDH2O
+
+
+print(sys.path)
+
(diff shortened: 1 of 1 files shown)


mdymczyk:
This is a good start but we'll need something a bit more automated - either assertions in the test (which might be tricky to get right, if there's a CPU/GPU hiccup a test might randomly fail) or a way to visualize the results (like Kibana, that everyone could see and check from time to time). Otherwise, we'll probably never look at those again.

I wanted to make something generic last week but am working now on improving the C++ testing framework so if anyone has some spare time that would be a great thing to implement.

pseudotensor:
make testperf is automated, just need to inject results into something lika kibana.  It's a github issue already.  Evwrything test we do should have a normal mode and a perf mode

navdeep-G:
There are assertions in the test now. I think initially it’s fine. We can always hook it up to some visualization. Maybe not Kibana, but something lightweight.

PR #254, opened 2017-10-31
Update Docker runtime files to use virtual env

navdeep-G:
* Utilize virtual env for Docker runtime images as these images are mainly used to set up environments for jupyter notebooks/demos.
* We can revisit pyenv later if it is absolutely needed.


pseudotensor:
Just don't expect anything build with virtualenv to work properly.  Runtime should be ok.  It seems silly to mix both pyenv and virtualenv.  I don't see the point.

navdeep-G:
Ya. Just a quick solution to get it working. Can try to get pyenv back if needed.

pseudotensor:
Yes, more important to get things working, totally agreed, thanks for fixing things.  We can figure out simple solution later.  I barely understand how this stuff works only lately.

Issue #161, opened 2017-09-12
Fallback to CPU/sklearn when GPU fails (e.g. out of memory)


pseudotensor:
And, fallback to CPU or sklearn if n_gpu=0 is set and we have no CPU algorithm except scikit-learn's.  E.g, if CPU lib not found for kmeans, revert to sklearn.

PR #260, opened 2017-11-07
#161 KMeans CPU fallback

mdymczyk:
#161 fall back to scikit CPU kmeans implementation when no GPU selected/present.


mdymczyk:
@pseudotensor yes all thanks to you explicitly creating `model_h2o4gpu` and `model_sklearn` beforehand. Another added bonus of that is that in DAI we can do `model.sklearn = OTHER_OBEJCT` since for example in DAI they are using MiniBatchKMeans and not KMeans.

PR #265, opened 2017-11-08
Fixes #256 kmeans tolerance

### LINKED ISSUE #256: KMeans threshold stopping criteria wrong
The threshold criteria should be checked against the total distance change between all old and updated centroids, not against how many centroids have been moved.

The current implementation can be going back and forth in a loop changing each centroid only by a tiny distance back and forth.

mdymczyk:
Previous impl was checking the % of points that got relabeled at each iteration. This is a bad tolerance metric as we can be doing a lot of tiny movements (sometimes even back and forth) all the time. Instead checking the total distance by which centroids got moved is a better metric

This improves the runtime in some cases, as we will not be doing unnecessarily many iterations.

### DIFF
--- a/src/gpu/kmeans/kmeans_impl.h
+++ b/src/gpu/kmeans/kmeans_impl.h
@@ -9,6 +9,7 @@
 #include <sstream>
 #include <thrust/device_vector.h>
 #include <thrust/reduce.h>
+#include <thrust/inner_product.h>
 #include "kmeans_centroids.h"
 #include "kmeans_labels.h"
 #include "kmeans_general.h"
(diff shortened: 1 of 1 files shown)


src/gpu/kmeans/kmeans_impl.h
```diff
@@ -212,34 +217,21 @@ int kmeans(
...
+      thrust::device_vector<T> diff(d_old_centroids.size());
+
+      thrust::transform(
+          d_old_centroids.begin(), d_old_centroids.end(),
+          (*centroids[0]).begin(),
+          diff.begin(),
+          thrust::minus<T>()
+      );
+      thrust::device_vector<T> squared_norm(1);
+      kmeans::detail::make_self_dots(1, k * d, diff, squared_norm);
```
RAMitchell: Consider using thrust inner product instead of make self dots: https://thrust.github.io/doc/group__transformed__reductions.html

This can actually be fused with the previous transform operation by defining a binary operator as (a-b)^2. Then you don't need the diff array. 

Remind me to also show you how to define a custom allocator for thrust at some point. This is important for performance reasons when calling certain thrust functions in a loop e.g. (reduction, inner product).

mdymczyk: @RAMitchell sweet, thanks!

Are you coming to MV for H2O World? Could start making a list of tips-and-tricks you could show us live then :-)

PR #268, opened 2017-11-09
Add mr-dl16 to jenkins

pseudotensor:


mdymczyk:
Maybe we could instead use `!mr-dl8` to check any new `nvidia-docker` machines as they get added?

pseudotensor:
Maybe, but mr-dl10 isn't setup right either.  Long term there shouldn't be any specific machines listed.

Issue #279, opened 2017-11-13
Improve KMeans memory footprint


mdymczyk:
Currently, KMeans requires vasts amounts of memory due to distance matrix calculation. For example, a 5,000,000 record dataset of floats clustered into 256 clusters will take at least 5.12GB memory for that matrix alone. When running `k-means++` this will be even bigger as it might choose more potential centroids.

Two options:

1) check if calculating the whole matrix is required or can we just multiply data_dot_product times centroid_dot_product and use that like Spark does https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/mllib/clustering/KMeans.scala#L559

Needs a check for numerical stability.

2) if the above is not a feasible solution then we can batch those computations into smaller chunks if not enough memory is available on the device.

mdymczyk:
For the time being, implemented 2), each `pairwise_distances` computation is batched should there be insufficient memory on the GPU.

Issue #278, opened 2017-11-13
Memory leak in KMeans


mdymczyk:
KMeans fails (every time) with a segfault after it finishes computing for `c=7` (so no computations for `c=[2,3,4]` are performed):

```
from h2o4gpu import KMeans
import numpy as np
from sklearn.datasets.samples_generator import make_blobs

samples = list(reversed([10000, 50000, 250000, 500000, 1000000]))
c = list(reversed([2,3,4,7,8,15,16,31,32,63,64,100]))
features = list(reversed([2, 4, 7, 8, 16, 32]))

for centers in c:
	for n_samples in samples:
    	for n_features in features:
			for i in range(10):
            	X, true_labels = make_blobs(n_samples=n_samples, centers=centers, cluster_std=1., n_features=n_features)
                X = X.astype(np.float32)
                kmeans_h2o = KMeans(n_gpus=1, n_clusters=centers).fit(X)
```

Running all of those scenarios one by one works fine.

PR #280, opened 2017-11-13
Fixes #194 #278 #279 and perf improvements

### LINKED ISSUE #194: Remove init_data from KMeans
While adding `_ptr` methods to KMeans I noticed that `init_data` isn't really necessary.

On a single GPU, it's not used at all as the data is sorted by label anyway in the algo.

In a multi-GPU setting the distribution of data (whether it's random or we just put consecutive batches of data on GPUS) shouldn't really matter as in the end on each GPU we are just summing them up and counting the number of data points for each centroid. Then we do the averaging on the host. Both `selectstrat` and `randomselect` should yield the same result.

Am I missing something?

@pseudotensor

mdymczyk:
* Remove init_data - the algorithm does not rely on data order but shuffling the initial data has a performance overhead (needs to copy data in a loop instead of a single memcpy call, needs to sort the output labels into the original order)
* Fixes bug with selectstrat - not all data was copied to the device
* Don't use fromiter for ctype pointers when casting back in python - very slow
* Use thrust vector contructor to copy data instead of running it in a loop
* Don't use type T for weights in kmeans++, takes unecessary space
* Use addAtomic for int when counting weights instead of for float/double - much faster
* Pairwise distance calculation is done in a batched manner if there's not enough memory on the GPU - now we are mainly limited by `rows * cols` size.

Batched pairwise computation code is still a bit copy pasted here and there because `nvcc` doesn't let me to put a `__device__` lambda inside a regular lambda...

### DIFF
--- a/examples/cpp/kmeans_driver.cpp
+++ b/examples/cpp/kmeans_driver.cpp
@@ -51,14 +51,13 @@ int main(int argc, char **argv) {
   int verbose=1;
   const char ord='r';
   int init_from_data=1;//true as set above, otherwise internally will set initial centroids "smartly"
-  int init_data=2; // randomly (without replacement) select from input
   int seed=12345;
   int dopredict = 0;
 
   std::vector<real_t> centroids(1);
   centroids[0] = static_cast<real_t>(drand48());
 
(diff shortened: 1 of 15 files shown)


mdymczyk:
2 more fixes necessary:

- multi-gpu bugs, some centroids are not calculated correctly
- in k-means|| we need to add distance calculation batching, this operation kills the GPU RAM

Issue #277, opened 2017-11-11
More Jenkins todo list (please contribute to list and start to prioritize)


pseudotensor:
~1) versioning~

~2) Add containers wheel cuda8/cuda9 links on README.  Update links to cuda8/cuda9 etc. wheels.~

~Do this automatically instead of hard-coding.~

3) have sklearn as import instead of complicated submodule (so no longer compile sklearn).  Wheel should then be more general, although still x86 due to C backend.

~4) name artifacts correctly and well~

https://github.com/h2oai/h2o4gpu/pull/291

~5a) Loop over builds as functions instead of copy/pasted stuff.~

~https://stackoverflow.com/questions/42158539/how-to-define-and-use-function-inside-jenkins-pipeline-config~
~https://github.com/h2oai/xgboost/blob/4d8433f145b4f3da0c60c69a9b0c9be8e1e52dd0/Jenkinsfile~

~5b) Parallel build for all cuda~ and python versions.  ~Even parallel on same machine (compile uses not too many cores most of the time)~

~https://stackoverflow.com/questions/43913698/jenkinsfile-parallel-directive~
~https://jenkins.io/doc/pipeline/examples/#jobs-in-parallel~

~NOTE: Can just have multiple Jenkinsfile(s) and pass into them the nccl/nonccl cuda8/cuda9 tags to control which is done from a Jenkins config in the browser.  Then the builds will be in full parallel mode more easily.  I'm not sure how this will show up in github/jenkins.~

~6) Have make smoke for build to do basic tests of other builds and containers (only do full make test once)~

~7) Have h2oai/xgboost repo in h2oai branch build wheel as separate jenkins pipeline and pushed to s3.  (This way don't have to build xgb everytime h2o4gpu built.)~

~https://github.com/h2oai/h2o4gpu/pull/291~
~https://github.com/h2oai/xgboost/commit/db585d1c04763d7f08eb1fec59e03c72faeefd36~

8) make testperf as benchmark if triggered by (say) #benchmark in commit or if any filles in src are changed. (like DAI for h2oaicore files)

[...]

PR #290, opened 2017-11-14
Improved versioning #277

mdymczyk:
Version stored in one place (`version.mk`). Probably still something wrong with the jenkinsfile + need to make sure everything gets cleaned properly when necessary.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,9 @@ SHELL := /bin/bash # force avoidance of dash as shell
 CONFIG=make/config.mk
 include $(CONFIG)
 
+VERSION=make/version.mk
+include $(VERSION)
+
 # System specific stuff
 include src/config2.mk
 
@@ -53,6 +56,14 @@ DOCKER_VERSION_TAG ?= "latest"
 #
 S3_CMD_LINE := aws s3
 
+#
+# BUILD_INFO setup
+#
+H2O4GPU_COMMIT ?= $(shell git rev-parse HEAD)
+H2O4GPU_BUILD_DATE := $(shell date)
+H2O4GPU_BUILD ?= "LOCAL BUILD @ $(shell git rev-parse --short HEAD) build at $(H2O4GPU_BUILD_DATE)"
+H2O4GPU_SUFFIX ?= "+local_$(shell git describe --always --dirty)"
+
 help:
 	@echo "make                 fullinstall"
 	@echo "make fullinstalldev  Clean everything, then compile and install project for development."
@@ -96,7 +107,7 @@ cpp:
 c:
 	$(MAKE) -j all -C src/interface_c
 
-py: apply_sklearn_simple
+py: apply_sklearn_simple build/VERSION.txt
 	$(MAKE) -j all -C src/interface_py
 
 pylint:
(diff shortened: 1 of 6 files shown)


mdymczyk:
@mmalohlava nope, I just commit it in sometimes by accident (bad habit of using `-a` when doing git commit), thanks for pointing it out.

Still some fixes are necessary (think we're not cleaning everything properly) but so far so good 👍

mdymczyk:
Ah also not sure how to link now from README to the latest bleeding-edge whl

mmalohlava:
`s3up` umi udelat latest link

Makefile
```diff
@@ -322,17 +322,17 @@ apply_sklearn_initmerge:
...
 
-buildjenkins: update_submodule cpp c py
+buildjenkins: update_submodule cpp c
 
-installjenkins: pyinstall
+buildpy: py pyinstall
 
 # for nccl cuda8 build
-fullinstalljenkins: cleanjenkins alldeps_private buildjenkins installjenkins
+buildjenkins: cleanjenkins alldeps_private buildjenkins
```
mmalohlava: In general, i would recommend to keep standard execution paths, if there is a separated execution path for building at jenkins, then it can easily miss problems which are visible by developers.

mdymczyk: @mmalohlava yes, the Makefile is due for some cleanup now that we don't use pydatatables and probably can be simplified

mdymczyk:
@mmalohlava think this should pass and work as intended. Any obvious suggestions I should fix?

Two things:

1) in other projects I noticed the Jenkinsfile/gradle version is subbing (with `tr`) all `+` signs - not sure why that was done so I left it as `+` everywhere for now

2) still not sure how to point from the README to the latest version? Yes `VERSION.latest` bucket is getting updated but the file contains the build number - different every time. Know any way to go around this? I tried setting the whole S3 bucket to public so we could point users there but that doesn't seem to work? Other thing I'm considering is to remove the version number from the wheel file name?

pseudotensor:
Looks like what you are doing heavily conflicts with what I've done in the other PR.  Any idea how to proceed?  Are all of your changes real, or did you do alot of reformatting?

mmalohlava:
@mdymczyk regarding 1. question - not necessary, i think i did that just to avoid having `+` in version name to avoid any later collisions 

regarding 2.question - it depends what you would like to achieve? Do you want to have link to the latest nightly/latest release? My idea is to always upload to private space (artifacts.h2o.ai) but then have additional step for promotion - for nightly it could be automatic, for releases manual click. The promotion will just take private artifacts and put them into public space and updates all links.

src/interface_py/Makefile
```diff
@@ -70,6 +70,8 @@ install:
 #	python3 setup.py sdist bdist_wheel
 
 clean:
+	rm -f h2o4gpu/__init__.py
```
mmalohlava: Are you sure here?

mdymczyk: yes, this is because how we bundle other libs with ours (sklearn, xgboost etc) - this file gets generated during make and if I clean the dependencies then I cannot load the module because the old init.py still refers to them

mdymczyk:
@mmalohlava in the readme we have a link to the bleeding edge build, which is a static link and worked until now because each build generated an artifact with the same name, i.e. `s3://bucket/0.0.4/h2o4gpu-0.0.4.py` (which was also put in `0.0.latest`). Now the problem is we'll put into the `0.1.0` and `0.1.latest` buckets files with the whole version, i.e. X.Y.Z+local_version. Not sure how to link to that from our main readme.

I.e. I will run a build and the artifact will be `s3://bucket/0.1.latest/h2o4gpu-0.1.0+master.1.py` but then after the next build it will be `s3://bucket/0.1.latest/h2o4gpu-0.1.0+master.2.py`. I was hoping I could make a symlink in S3 like `s3://bucket/0.1.latest/h2o4gpu-latest.py` which I could modify after s3up to point to the newly uploaded artifact (then I could just hard code in the readme that as the URL to download) but doesn't seem S3 supports that...

mdymczyk:
Implemented in another PR.

PR #291, opened 2017-11-14
Update Jenkins with s3 retry, 1 Dockerfile-build and 1 Dockerfile-runtime for both cuda8 and cuda9, and add data to docker runtime necessary for some selected demos

pseudotensor:


pseudotensor:
Mostly small limited few data sets.  Should take about 10 seconds to download.  Can use s3 instead of wget or use local data sources if already synched.

Makefile
```diff
@@ -129,14 +129,14 @@ fullinstall2: clean alldeps2 sync_open_data build install
 	mkdir -p src/interface_py/dist2/ && cp -a src/interface_py/dist/*.whl src/interface_py/dist2/
 
 runtime:
-	$(MAKE) py
+	$(MAKE) fullinstall
```
navdeep-G: Why a fullinstall?

Makefile
```diff
@@ -129,14 +129,14 @@ fullinstall2: clean alldeps2 sync_open_data build install
...
 	@echo "+--Building Runtime Docker Image--+"
-	docker build -t opsh2oai/h2o4gpu-runtime:latest -f Dockerfile-runtime .
+	nvidia-docker build -t opsh2oai/h2o4gpu-runtime:latest -f Dockerfile-runtime .
 
 runtimecuda9:
-	$(MAKE) py
+	$(MAKE) fullinstall
```
navdeep-G: Same as before.

Makefile
pseudotensor: We talked about this together before.  The local system won't have the correct C++ backend, xgboost, etc. unless one does make fullinstall.  E.g., if do only make py, then if fullinstall previously was for cuda8, then make py will not be enough to get the cuda9 binary.

Makefile
pseudotensor: This should be make fullinstall2

navdeep-G:
Please squash these on merge.

Makefile
```diff
@@ -129,14 +129,17 @@ fullinstall2: clean alldeps2 sync_open_data build install
...
-
+	$(MAKE) fullinstall
+	nvidia-docker build -t opsh2oai/h2o4gpu-cuda8-runtime:latest -f Dockerfile-runtime --build-arg cuda=nvidia/cuda:8.0-cudnn5-devel-ubuntu16.04 .
+	nvidia-docker save opsh2oai/h2o4gpu-cuda8-runtime > h2o4gpu-cuda8-runtime.tar
+	gzip  h2o4gpu-cuda8-runtime.tar
 runtimecuda9:
-	$(MAKE) py
 	@echo "+--Building Runtime Docker Image for cuda9--+"
-	docker build -t opsh2oai/h2o4gpu-cuda9-runtime:latest -f Dockerfile-cuda9-runtime .
+	$(MAKE) fullinstall2
```
navdeep-G: A better name might be `fullinstallcuda9`?

Makefile
```diff
@@ -316,6 +319,9 @@ fullinstalljenkins: cleanjenkins alldeps_private buildjenkins installjenkins
 # for nonccl cuda8 build
 fullinstalljenkins2: cleanjenkins alldeps_private2 buildjenkins installjenkins
 	mkdir -p src/interface_py/dist2/ && mv src/interface_py/dist/*.whl src/interface_py/dist2/
+# for nccl cuda9 build
+fullinstalljenkins4: cleanjenkins alldeps_private buildjenkins installjenkins
+	mkdir -p src/interface_py/dist4/ && mv src/interface_py/dist/*.whl src/interface_py/dist4/
 # for nonccl cuda9 build
 fullinstalljenkins3: cleanjenkins alldeps_private2 buildjenkins installjenkins
 	mkdir -p src/interface_py/dist3/ && mv src/interface_py/dist/*.whl src/interface_py/dist3/
```
navdeep-G: What does `fullinstalljenkins*` imply in terms of the number system used?

Makefile
```diff
@@ -129,14 +129,17 @@ fullinstall2: clean alldeps2 sync_open_data build install
 	mkdir -p src/interface_py/dist2/ && cp -a src/interface_py/dist/*.whl src/interface_py/dist2/
 
 runtime:
-	$(MAKE) py
 	@echo "+--Building Runtime Docker Image--+"
-	docker build -t opsh2oai/h2o4gpu-runtime:latest -f Dockerfile-runtime .
-
+	$(MAKE) fullinstall
```
navdeep-G: I would call it `fullinstallcuda8` just so we know whats exactly going on.

Makefile
pseudotensor: Ya, all the names can be changed once things actually work.  Note that the names don't ensure anything, since what was done previously to this PR had names wrong (there was a cuda8 instead of a cuda9).  Having a loop over stages and select from a list is best, but someone else can pick that up.

pseudotensor: Also, alot of this going to be deprecated by my changes to the h2oai branch of our xgboost.  It already uploads 4 xgboost wheels, so we don't need all 4 cases for fullinstalljenkins.

PR #293, opened 2017-11-16
Update how jenkins artifacts named and use simple s3 upload

pseudotensor:


mdymczyk:
@pseudotensor Jon, are all these changes really necessary *now*? I thought we agreed to just use the simple script for s3up so master gets green, I'd then apply my versioning branch and then we'd redo the Jenkins file the way you're doing here. These are major changes and it will make it just so much harder for me to merge and is still blocking my merge.

PR #294, opened 2017-11-17
[WIP] Improved artefact versioning

mdymczyk:
* Applied from the original PR due to merge conflicts
* Pulled out most duplicated code into methods in Jenkinsfile

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -466,7 +466,6 @@ pipeline {
     }
 }
 
-@NonCPS
 void publishToS3(String extratag, String dist) {
     def versionTag = buildInfo.get().getVersion()
     def artifactId = "h2o4gpu"
@@ -489,7 +488,6 @@ void publishToS3(String extratag, String dist) {
     //}
 }
 
-@NonCPS
 void publishRuntimeToS3(String extratag) {
     def versionTag = buildInfo.get().getVersion()
     def artifactId = "h2o4gpu"
@@ -512,7 +510,6 @@ void publishRuntimeToS3(String extratag) {
     //}
 }
 
-@NonCPS
 void runTests(String dockerimage, String extratag, String dist, String target) {
     def versionTag = buildInfo.get().getVersion()
 
@@ -536,7 +533,6 @@ void runTests(String dockerimage, String extratag, String dist, String target) {
     }
 }
 
-@NonCPS
 void buildOnLinux(String dockerimage, String extratag, String dist) {
     withCredentials([[$class: 'AmazonWebServicesCredentialsBinding', credentialsId: "awsArtifactsUploader"]]) {
         echo "Building on linux - running docker"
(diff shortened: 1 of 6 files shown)


navdeep-G:
Still some conflicting files it seems like? ;(

mdymczyk:
@navdeep-G just renamed some "dist" to "dist2" etc. in master to make the build pass (but turns out there were other issues anyway), nothing major. If this PR will work I will be able to just use this version as is. Please, make sure no one at the office works on the Jenkinsfile today.

Jenkinsfile
```diff
@@ -1046,7 +465,132 @@ pipeline {
     }
 }
 
+@NonCPS
```
navdeep-G: Love adding functions and simplifying the build stages.

src/interface_py/setup.py
```diff
@@ -108,9 +95,13 @@ class BinaryDistribution(Distribution):
...
+# Read version
+about_info={}
+with open('__about__.py') as f: exec(f.read(), about_info)
+
 setup(
     name='h2o4gpu',
-    version='0.0.4',
+    version=about_info['__version__'],
     author='H2O.ai, Inc.',
     author_email='h2ostream@googlegroups.com',
```
navdeep-G: Not sure if this link is still useful. We can check with @ledell to see whats a proper link for support. Not important now,but if we go to pypi then it would be more important.

src/interface_py/Makefile
```diff
@@ -70,6 +70,12 @@ install:
 #	python3 setup.py sdist bdist_wheel
 
 clean:
+	rm -f h2o4gpu/BUILD_INFO.txt
+	rm -f h2o4gpu/dist*
+    # A bit hacky b/c of how we are merging __init__.py with other frameworks
+    # This *is* required
+	rm -f h2o4gpu/__init__.py
```
navdeep-G: I wonder if there is a better solution? I think its fine for now, but just curious.

pseudotensor: In works (as in jenkins list issue and as separate issue) is to pull out sklearn completely.  That will fix build times, this, etc.

mdymczyk:
@pseudotensor yeah the snapshots are being uploaded and the docker runtime is trying to use them - for some reason it fails with a 403, looking into why.

pseudotensor:
403 just means file not found really.  Is that your + thing?

mdymczyk:
No, 403 is auth failed - you'll get the same error if you try to just copy paste the url and go there using a browser. All works with %2B instead of +. Also tried on dl14.

pseudotensor:
yes i am saying auth failed is given even when file does not exist

mdymczyk:
@pseudotensor ah ok, sorry I'm getting slow :-) Yeah that's the `+` thing - pushed a fix.

pseudotensor:
What about updating the README links?  Easy to do later of course.

mdymczyk:
@pseudotensor yes will do that too, but I need to add an upload that will upload an artifact without the build number for that and to a bucket without build number, so we the URL is the same every time (now the artifact is something like `h2o4gpu-0.1.0+master.1.whl` etc. need to upload one without the `+master.1` part). But might need to grab some sleep first.

PR #295, opened 2017-11-18
Clean-up jenkins so only 1 file with separate functions.

pseudotensor:


mdymczyk:
Looks nice, 2 things so far:

1) think we can define `def utils` at the top of each file like we do with `utilsLib`, so we don't have to define it 5 times in each file
2) the container names for each file will have to be changed so each is unique or we'll have collisions (we can just add `cuda8-nccl` etc. to each)

will add those if I'll have time today

mdymczyk:
Pushed some changes but not sure they will work:

1) `BUILD_TYPE` I added in each file so the container names dont collide - I don't know if it will be accessible in the util file? Might need to pass it there differently?

2) Not sure if the `buildInfo` used in methods can be initlized in the actual Jenkinsfiles and used implicitly in our util methods. If not we just need to add `BuildInfo buildInfo` parameter to every method and pass it to it.

mdymczyk:
Right, read a bit more about loading external scripts in Jenkins - it does seem very limited and a simple `def h2o4gpuUtils = load 'path'` won't work. What I don't understand is why this also doesn't work:

```
def h2o4gpuUtils() {  return load 'path' }

stage {
    h2o4gpuUtils().ourMethod()
}
```

As for CHANGE_ID, the documentation is a bit useless:

> CHANGE_ID
> An identifier corresponding to some kind of change request, such as a pull request number

So not sure what exactly it is but I noticed in previous version the build was complaining that the container with a given name already exists.

Anyway it seems the current state works so lets leave it as it is and improve, if needed, incrementaly later on. 

Anyway good job us :-)

mdymczyk:
One more thing I want to change, and need some feedback before I do so. I'd like the uploads to be:

VERSION = 0.1
VERSION_NO_LOCAL = 0.1.0
FULL_VERSION = 0.1.0+branch.build

1) Snaphosts go to `artifacts.h2o.ai/snapshots/ai/h2o/h2o4gpu/VERSION/h2o4gpu-FULL_VERSION.whl`
2) Bleeding edge goes to `artifacts.h2o.ai/releases/bleeding-edge/ai/h2o/h2o4gpu/VERSION/h2o4gpu-VERSION_NO_LOCAL.whl`
3) Bleeding edge goes to `artifacts.h2o.ai/releases/stable/ai/h2o/h2o4gpu/VERSION/h2o4gpu-VERSION_NO_LOCAL.whl`

How does this sound?

pseudotensor:
Getting no space left on device on jenkins, but otherwise script running: http://mr-0xc1:8080/blue/organizations/jenkins/h2o4gpu%20nonccl%20cuda9/detail/PR-295/22/pipeline

pseudotensor:
I assume you meant the 3rd item to be stable, not bleeding edge.  So you want to remove Z in X.Y.Z for stable?  I guess so.  What's the need?  Is that normal?

pseudotensor:
Jenkins is currently messed up with hangs and out of space errors.  If I try to abort a job nothing happens.  I've told @mmalohlava and @abal5 .  I won't work on the updates until these issues are figured out.  I'll probably only work on pulling  some jenkins commands into make commands like is going on with DAI.  That way can run locally to build things, like runtime that I need urgently.

mdymczyk:
@pseudotensor yes I also noticed the hangs, maybe it has to do something with increased number of executors?

As for versioning, I was thinking of removing `Z` for all artifacts but only for the bucket name, the wheel file would be either `X.Y.Z` or `X.Y.Z+branch.build`. In such a case we'd have all the bleeding edge/stable wheels in a `X.Y` folder. Or we can put them in `X.Y.Z` both are ok. Since now we're creating a new folder for each build.

pseudotensor:
Sure, on X.Y.Z, sounds fine.  Whatever you think.

pseudotensor:
FYI, I'd also like to add to this PR the pull-out of docker commands into scripts that either the Jenkins or a local make would be able to use.  This is what DAI is also doing.  Ensures a local run is same as jenkins and that the docker and non-docker can be made the same.  So when test locally, most likely to be a valid test of docker and valid test of what jenkins does.

mdymczyk:
@pseudotensor 👍 but could we first squash and merge this PR and do it in a new one? So people trying out 4gpu could use newest builds.

pseudotensor:
We could, but I guess so far we are just in jenkins land and there are no actual new commits to the wheel.  Right?

mdymczyk:
Now no, but the master build has been failing for some time now and there's quite a lot of fixes to KMeans missing from the latest published bleeding edge.

pseudotensor:
Note that right now the runtime docker is incorrect (needed to commit updates).  So, best to not commit to master something that's not quite right.  Whatever I do it's still just part of cleaning up jenkins files and making sure runtime docker is built correctly.

mdymczyk:
But that part is also incorrect in master, no? So by merging this PR first we'd at least get the build green and new wheels out without really pushing any new bugs, no?

pseudotensor:
Either way is fine, but I don't see the need to rush this PR given it's been in limbo for a few days.  I should be finished with changes soon.

pseudotensor:
Ah, you are all green, so go ahead and squash and merge.  I'll make new PR for my stuff.

PR #296, opened 2017-11-21
Remove build details from Jenkins and put into Makefile for either user or jenkins to run

### LINKED ISSUE #299: Have runtime docker use local stash or file instead of s3
As for the runtime docker depending upon S3, I'm not sure what to do about that. I think I did that to avoid having to regenerate the wheel if it happens to already be on S3. So a user can just make the runtime docker without building the wheel. Also, the jenkins build will use the s3 file for the runtime docker (however, for that, seems we can just use stashed version). Feel free to fix this to just always be pip installing from the correct ${dist} path.
￼
￼  remember why I used s3 for the runtime docker. If one installs the wheel during "docker build", I had at the time thought one couldn't mount a path and so one couldn't access the needed wheel file. However, the build context (. at end of build) allows one to COPY from that context. So all one has to do is to add a COPY command that copies the src/interface_py/${dist}/*.whl to the local path and use that wheel when pip installing.
Since you are familiar with the versioning, you can probably best ensure the wheel used is the correct name, etc.
I fixed the dockerignore so src is in the build context again.

pseudotensor:
* Remove build details from Jenkins and put into Makefile for either user or jenkins to run
* Clean environment for docker to build or run without forcing working space to be in native disk.
* Docker copies local native disk to /repo/ inside docker to avoid contaminating / and to allow (e.g.) mrproper to properly clean without deleting docker system files
* Copy only needed artifacts from docker back into native disk
* Added .dockerignore to avoid building context with excessive files
* Fixed runtime docker build to commit data and save with data
* Fixes #299 
* Fixes most of #277


pseudotensor:
@mdymczyk 

I notice at the end of: http://mr-0xc1:8080/blue/rest/organizations/jenkins/pipelines/h2o4gpu%20nccl%20cuda8/branches/PR-296/runs/17/nodes/12/steps/43/log/?start=0

I see: 

```
running install_egg_info
running egg_info
writing h2o4gpu.egg-info/PKG-INFO
writing dependency_links to h2o4gpu.egg-info/dependency_links.txt
writing requirements to h2o4gpu.egg-info/requires.txt
writing top-level names to h2o4gpu.egg-info/top_level.txt
reading manifest file 'h2o4gpu.egg-info/SOURCES.txt'
writing manifest file 'h2o4gpu.egg-info/SOURCES.txt'
Copying h2o4gpu.egg-info to /root/.pyenv/versions/3.6.1/lib/python3.6/site-packages/h2o4gpu-0.1.0_local_-py3.6.egg-info
running install_scripts
make[1]: Leaving directory '/repo/src/interface_py'
make: *** No rule to make target '17'.  Stop.
Docker devel - Clean local wheels and Copying wheel from docker
```

pseudotensor:
Also, I think I messed-up your version stuff a bit.  The wheel files look like: "h2o4gpu-0.1.0_local_-py36-none-any.whl" as if the build number of hash is missing.

pseudotensor:
@mdymczyk Also, why are so many new bleeding edge directories made?

image

pseudotensor:
With the file names, the mixup seems to be a _ vs. a - as well as a missing build number:

```
[h2o4gpu_nccl_cuda9_PR-296-UVRC3DJXR7GDVCF5DXFNBIWSZLCHXSR375VHTMP7Z3PB2ELTGHBQ] Running shell script
+ echo Stashed wheel file:
Stashed wheel file:
+ ls -l src/interface_py/dist4/
total 51348
-rw-r--r-- 1 jenkins jenkins 52579078 Nov 21 00:15 h2o4gpu-0.1.0_local_-py36-none-any.whl
Running tests
— Print Message
<1s
Running tests
Shell Script
9m 0s
Publishing artifact to S3
— Print Message
<1s
Shell Script
<1s
[h2o4gpu_nccl_cuda9_PR-296-UVRC3DJXR7GDVCF5DXFNBIWSZLCHXSR375VHTMP7Z3PB2ELTGHBQ] Running shell script
+ s3cmd put src/interface_py/dist4/h2o4gpu-0.1.0+local_-py36-none-any.whl s3://artifacts.h2o.ai/snapshots/ai/h2o/h2o4gpu/0.1.0+local_-nccl-cuda9/
ERROR: Parameter problem: Nothing to upload.
script returned exit code 64
```

mdymczyk:
@pseudotensor regarding the directories, this is what I was trying to explain in the other PR - currently we are pushing data to a `${versionTag}${extratag}` bucket, the `versionTag` contains the full version i.e. also the "local part" (`+master.1` etc.) - when you're done with your changes here I want to change it so we push all the files to a `X.Y` bucket instead so we don't create so many folders.

pseudotensor:
@mdymczyk Ok, but as in my other message, the version stuff isn't working anymore.  Is it obvious to you why?  Can you fix the branch?

mdymczyk:
The wheel file name will have the form of `h2o4gpu-0.1.0_local_-py36-none-any.whl` if you use `_local` instead of `+local`, that `+` sign is important (part of Pythons versioning schema).

pseudotensor:
I didn't use _local on purpose.  Didn't touch it.

mdymczyk:
Ok, looking through the changes.

mdymczyk:
Unrelated, any reason we're removing: `def email = utilsLib.getCommandOutput("git --no-pager show -s --format='%ae'")`? I added it so outside contributors could have access to build failures, since Jenkins is not visible outside the company.

pseudotensor:
FYI, currently the try catch finally is commented out.

pseudotensor:
I removed the email thing because it wasn't working anymore.  It requires a git command and it failed.  I can try to find the commit where that failed.  Once things are working we can try to add it back again.

mdymczyk:
Ah ok fair enough!

pseudotensor:
The buildinfo issue and s3 upload issue is probably related to that "No rule to make target '17'.  Stop." error.  It seems that the H2O4GPU_BUILD and H2O4GPU_SUFFIX aren't making it down properly and the make line is as if it's trying to run make H2O4GPU_BUILD itself.

mdymczyk:
@pseudotensor if I got the syntax right it should be back to how it was, we have to pass `${H2O4GPU_BUILD}` like an env variable to `fullinstalljenkins` i.e. `H2O4GPU_BUILD=${H2O4GPU_BUILD}`, otherwise it gets lost since we're calling it inside docker.

pseudotensor:
@mdymczyk The commit just before the trycatchfinally stuff is green.  Any other ideas why the try catch stuff fails?

pseudotensor:
Sure, you mean in the Makefile.  In principle, the version is whatever the user wants it to be, and the runtime docker part still pulls from s3 so it needs to be a version that exists currently.  So that part is not perfect.

pseudotensor:
Feel free to fix up the rest of the things you mentioned regarding versioning (Makefile change and hash usage, etc.) unless you want to do that in a separate PR.

As for the runtime docker depending upon S3, I'm not sure what to do about that.  I think I did that to avoid having to regenerate the wheel if it happens to already be on S3.  So a user can just make the runtime docker without building the wheel.  Also, the jenkins build will use the s3 file for the runtime docker (however, for that, seems we can just use stashed version).  Feel free to fix this to just always be pip installing from the correct ${dist} path.

pseudotensor:
I remember why I used s3 for the runtime docker.  If one installs the wheel during "docker build", I had at the time thought one couldn't mount a path and so one couldn't access the needed wheel file.  However, the build context (. at end of build) allows one to COPY from that context.  So all one has to do is to add a COPY command that copies the src/interface_py/${dist}/*.whl to the local path and use that wheel when pip installing.

PR #305, opened 2017-11-22
Principal Components Analysis

mdymczyk:
Still work in progress, any comments welcome.


navdeep-G:
Noticed a lot of files have been moved around. Assuming you used `git mv ...` to preserver the history? Please confirm this is saved.

src/gpu/pca/pca.cu
```diff
@@ -0,0 +1,51 @@
...
+		//Take in X matrix and allocate for X^TX
+		tsvd::Matrix<float>X(_param.X_m, _param.X_n);
+		X.copy(_X);
+
+		tsvd::Matrix<float>XtX(_param.X_n, _param.X_n);
+
+		//create context
+		tsvd::DeviceContext context;
+
+        tsvd::normalize_columns(X, context);
```
navdeep-G: This is what I anticipated you using. However, there is a `cub` version. Have you looked into that?

src/gpu/tsvd/tsvd.cu
```diff
@@ -240,12 +239,17 @@ void calculate_u(const Matrix<float> &X, const Matrix<float> &Q, const Matrix<fl
...
  */
 void truncated_svd(const double* _X, double* _Q, double* _w, double* _U, double* _explained_variance, double* _explained_variance_ratio, params _param)
+{
+    Matrix<float>X(_param.X_m, _param.X_n);
+	X.copy(_X);
+    truncated_svd(_X, _Q, _w, _U, _explained_variance, _explained_variance_ratio, _param);
+}
+
+void truncated_svd_matrix(Matrix<float> X, double* _Q, double* _w, double* _U, double* _explained_variance, double* _explained_variance_ratio, params _param)
 {
```
navdeep-G: Why the above change and an additional function? I think tsvd.cu can remain unchanged for this addition?

navdeep-G:
Added some additional comments around the cuda code. Will look at Python more once we have some tests.

src/gpu/tsvd/tsvd.cu
mdymczyk: @navdeep-G in PCA I first create the Matrix, then do normalization on it and then pass it to svd, by that time I already have a Matrix object so I would prefer to pass that instead of `double* _X` which would create the X matrix again.

src/gpu/pca/pca.cu
mdymczyk: @navdeep-G no, will check if this one turns out to be a bottleneck

mdymczyk:
@navdeep-G yes the history is saved, though it's just 1 commit :-)

src/gpu/tsvd/tsvd.cu
navdeep-G: Ah, okay, makes sense.

pseudotensor: @mdymczyk Aren't you only required to center the matrix?  You don't have to normalize it because you aren't creating the covariance matrix -> PCA and instead using SVD.

navdeep-G: Yes, should be center -> svd, right?

pseudotensor: FYI, center just means take each column and remove the average so each column average is zero.  Normalization is much more expensive.

navdeep-G: Yes, exactly. So just need to get each column mean and subtract from each obs. i already have a util to get each column variance. Can take that and make a new one, `mean_iterator`.

navdeep-G: To get variance. So slight modification for mean. 
```
template<typename T>
class variance_iterator{
public:
    // Required iterator traits
    typedef variance_iterator<T>          self_type;            ///< My own type
    typedef size_t                            difference_type;  ///< Type to express the result of subtracting one iterator from another
    typedef T                           value_type;             ///< The type of the element the iterator can point to
    typedef T*                          pointer;                ///< The type of a pointer to an element the iterator can point to
    typedef T                           reference;              ///< The type of a reference to an element the iterator can point to
    typedef std::random_access_iterator_tag     iterator_category;      ///< The iterator category
	const T* data_ptr;
	const T* mean_ptr;
	const int col_rows;
	size_t offset;
	__device__ T operator[](size_t idx){
		idx = idx + offset;
		T mean = mean_ptr[idx/col_rows];
		T dev_square = pow((data_ptr[idx] - mean),2);
		return dev_square;
	}
	__device__ self_type operator+(size_t idx){
		self_type retval(data_ptr, mean_ptr, col_rows);
		retval.offset += idx;
		return retval;
	}

	__host__ __device__ variance_iterator(const T* data_ptr, const T* mean_ptr, const int col_rows):data_ptr(data_ptr), mean_ptr(mean_ptr), col_rows(col_rows), offset(0){

	}
};
```

PR #308, opened 2017-11-22
svd test fix

pseudotensor:
Hotfix for broken master so no errors when running svd test. (fix for accidental pushes to master)


navdeep-G:
LGTM

PR #309, opened 2017-11-22
Fix make testperf so results generated even if some asserts hit.

pseudotensor:


pseudotensor:
FYI, this isn't tested, so I'm just going to squash and merge.  But just FYI that now the bash tests_open/showresults.sh will run because the "-" on the previous line means don't fail make if hit error code.  Error code is ok for testperf because it means h2o4gpu slower than h2o-3.  That's ok.

PR #310, opened 2017-11-22
Sync

pseudotensor:
Update make testbigperf and synch folder so correct commands used to get data.  Still, data is closed.


pseudotensor:
Just FYI, going to squash and merge since not tested in jenkins.

PR #327, opened 2017-12-02
KMeans color quant demo

mdymczyk:


pseudotensor:
Last thing needed is I run full notebook for last sklearn part -- will run while I sleep an push in morning and then check docker.

PR #337, opened 2017-12-04
Bump version to 0.2.0

mdymczyk:


mmalohlava:
4 files to change version? it looks suspicious...

mdymczyk:
@mmalohlava yes, `config/version.mk`, `README`, `Makefile` and one kmeans class, which doesnt even need it - will remove it.

I made an issue for `Makefile` - it should read the version from `version.mk`.

But `version.mk` and `README` will have to be changed every time. Unless we know a way to read the version from `version.mk` in `README`?

mmalohlava:
@mdymczyk hmm, it could be tricky - you can invoke JS which will access the file via HTTP and read it. Not worth it. But you can look if there is some kind of way to do that via image service:

image

mdymczyk:
@mmalohlava will have a look. For now I fixed the Makefile to read it from `make/version.mk` - from now on we'll only have to change `make/version.mk` and `README.md`.

Issue #334, opened 2017-12-03
KMeans: Multi-GPU performance saturated even with init='random'


pseudotensor:
10-15s with 1 or 2 GPUs on 1080ti for homesite kmeans notebook.  As if (compared to previously) improvements allowed for across the board for all GPU counts, but saturated in performance for a different reason that was possibly related to why 4 and 8 GPUs had similar performance before.

mdymczyk:
One issue I noticed is even if we set `random_state=1` in both cases (1 and 2 gpu), the total number of iteration differs. 2 GPU version is running for over 200iter, while 1 GPU version finishes around 150. Both cases are reproducible (i.e. 1GPU always runs for ~150, 2GPU always ~200) but they should be both similar. The only reason why this would happen I can come up with on the spot is because in 1GPU case we do scaling on the GPU while in the 2GPU version on the CPU.

pseudotensor:
On volta I get very poor performance with the kmeans image notebook.

GPU#   Time for fit
1            44s
2            29s
4            24s
8            30s # after warmup
8            45s  # first time run on gpus

The first time ran, it seems to be sequentially allocating on the GPU, as if you have some cuda operation (e.g. sync or info getter) that is sequential and getting the cuda context setup sequentially).  Need to avoid that.

For post-warmup, see low GPU usage for 8 GPUs:

![image](https://user-images.githubusercontent.com/2249614/33901172-cc0d7d6a-df25-11e7-8c10-1b64fd33a252.png)

pseudotensor:
Even larger works, although slow.

![image](https://user-images.githubusercontent.com/2249614/38060274-e8b1d340-329e-11e8-8b29-b0d5cb5c5a5f.png)

spends most of the time on the CPU doing single-threaded work.  Pokes on the GPU (1 at a time) sometimes.

Only after a while does it do use all GPUs in parallel:

![image](https://user-images.githubusercontent.com/2249614/38060353-2713b0ea-329f-11e8-8596-09ded3e5b48b.png)

Then it goes back to single-threaded stuff on CPU with no GPU usage.

![image](https://user-images.githubusercontent.com/2249614/38060358-349b240a-329f-11e8-885d-e28f8291eb73.png)

pseudotensor:
```
import h2o4gpu
import numpy as np
>>> X = np.random.rand(80636906,34)                               
>>> model = h2o4gpu.KMeans(n_clusters=5,random_state=1234).fit(X) 
[...]

PR #339, opened 2017-12-04
Fixes #333 KMeans++ perf.

### LINKED ISSUE #333: Kmeans: KMeans|| too slow compared to init='random'
On default homesite notebook, takes 1:10 on single 1080ti with kmeans||, but only 10s with init_random.  CPU/GPU usage shows hits GPU at first, but spends most of the time on the CPU, and does a bit more on the GPU at the end.

mdymczyk:
* Don't allocate new vectors at each iteration
* Don't erase from vectors (instead set weight to 0)

Haven't run on homesite yet but on a similarly sized dataset now I'm getting (before would run around 1m 15-30sec):

```
>>> n_samples = 260000
>>> centers = 1000
>>> X, true_labels = make_blobs(n_samples=n_samples, centers=centers, n_features=300)
>>> s = time.time()
>>> kmeans_h2o = KMeans(n_gpus=1, n_clusters=centers).fit(X)
>>> print(time.time() - s)
19.270898818969727
>>> v_measure_score(kmeans_h2o.labels_, true_labels)
0.98429668669984283
>>> s = time.time()
>>> kmeans_h2o2 = KMeans(n_gpus=1, init='random', n_clusters=centers).fit(X)
>>> print(time.time() - s)
10.995243072509766
>>> v_measure_score(kmeans_h2o2.labels_, true_labels)
0.9653079939768735
```

~~edit: for homesite it runs a lot more iterations than `init=random`, need to double check this change~~

Ran on homesite with `max_iter=1000, n_gpus=1` on my 1080 got `Wall time: 22.3 s`

#334 still present, though. For some values of `random_state` n_gpus=2 is actually slower than 1.

### DIFF
--- a/src/gpu/kmeans/kmeans_h2o4gpu.cu
+++ b/src/gpu/kmeans/kmeans_h2o4gpu.cu
@@ -239,7 +239,7 @@ template<typename T>
 void add_centroid(int idx, int cols,
                   thrust::host_vector<T> &data,
                   thrust::host_vector<T> &weights,
-                  thrust::host_vector<T> &centroids) {
+                  std::vector<T> &centroids) {
   for (int i = 0; i < cols; i++) {
     centroids.push_back(data[idx * cols + i]);
   }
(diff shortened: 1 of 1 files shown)


mdymczyk:
In the long run, I can see if running that part on the GPU would be beneficial, also, but this can work as a fast fix for now.

PR #363, opened 2017-12-14
Fixes #362

### LINKED ISSUE #362: KMeans calculate centroids kernel bug
The `calculate_centroids` kernel from `kmeans_centroids.h` is non-deterministic! The method basically goes through all the records and sums them up by labels, this summed records will be used to calculate the new centroids (by dividing them by number of rows for each label). This should return exactly the same value for same input but returns slightly different values.

```
model11 = h2o4gpu.KMeans(n_gpus=1, n_clusters=k, random_state=1, max_iter=1).fit(X)
model12 = h2o4gpu.KMeans(n_gpus=1, n_clusters=k, random_state=1, max_iter=1).fit(X)

all(model11.cluster_centers_ == model12.cluster_centers_)
``` 

Returns `False` for certain datasets, e.g. Homesite. For small `max_iter` it's ok but for full runs of the algorithm, it's pretty bad. Even worse when we compare runs on 1 and multiple gpus.

mdymczyk:
For certain datasets (i.e. Homesite) `atomiAdd(float*, float)` was *very* unstable.
This is due to the fact that, because of rounding errors, it is not a transitive operation
and we are computing new centroids using kernels which run in a parallel (indeterministic) fashion.

This fix uses Kahan summation, which works well but bumps the memory footprint and is here as a 'quick' fix. In the long term, we can change the centroids matrix to always be of type double.
This change will require a little bit more time, though.

Another thing to have in mind should we hardcode `centroids` to double is potential time perf hit - running Homesite using `double` is several times slower - bumping `centroids` to double might cause similar speed problems potentially?

EDIT: actually, instead of using Kahan, which seems to have some problems for multi-GPU, we can accumulate values here into `double`.

### DIFF
--- a/src/gpu/kmeans/kmeans_centroids.h
+++ b/src/gpu/kmeans/kmeans_centroids.h
@@ -84,9 +84,9 @@ __global__ void calculate_centroids(int n, int d, int k,
 
 template<typename T>
 __global__ void revert_zeroed_centroids(int d, int k,
-                                    T *tmp_centroids,
-                                    T *centroids,
-                                    int *counts) {
+                                        T *tmp_centroids,
+                                        T *centroids,
+                                        int *counts) {
   int global_id_x = threadIdx.x + blockIdx.x * blockDim.x;
   int global_id_y = threadIdx.y + blockIdx.y * blockDim.y;
   if ((global_id_x < d) && (global_id_y < k)) {
@@ -159,6 +159,10 @@ void find_centroids(int q, int n, int d, int k,
   int n_threads_y = 16; // TODO FIXME
   //XXX Number of blocks here is hard coded at 30
   //This should be taken care of more thoughtfully.
+
+  // TODO necessary?
+  cudaDeviceSynchronize();
+
   calculate_centroids << < dim3(1, 30), dim3(n_threads_x, n_threads_y), // TODO FIXME
       0, cuda_stream[dev_num] >> >
       (n, d, k,
@@ -168,9 +172,12 @@ void find_centroids(int q, int n, int d, int k,
           thrust::raw_pointer_cast(centroids.data()),
           thrust::raw_pointer_cast(counts.data()));
 
-  #if(CHECK)
+  // TODO necessary?
+  cudaDeviceSynchronize();
+
+#if(CHECK)
   gpuErrchk(cudaGetLastError());
-  #endif
+#endif
 
   // Scaling should take place on the GPU if n_gpus=1 so we don't
   // move centroids and counts between host and device all the time for nothing
@@ -184,9 +191,9 @@ void find_centroids(int q, int n, int d, int k,
             thrust::raw_pointer_cast(centroids.data()),
             thrust::raw_pointer_cast(counts.data()));
 
-    #if(CHECK)
+#if(CHECK)
     gpuErrchk(cudaGetLastError());
-    #endif
+#endif
 
     //Averages the centroids
     scale_centroids << < dim3((d - 1) / 32 + 1, (k - 1) / 32 + 1), dim3(32, 32), // TODO FIXME
--- a/src/gpu/kmeans/kmeans_centroids.h
+++ b/src/gpu/kmeans/kmeans_centroids.h
@@ -33,10 +33,9 @@ void update_centroid(int label, int dimension, int d,
                      T accumulator, T *centroids,
                      int count, int *counts) {
   int index = label * d + dimension;
-  T *target = centroids + index;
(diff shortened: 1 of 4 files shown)


mdymczyk:
Single GPU scenario seems fixed, multi-GPU still seems to fluctuate a bit from time to time in terms of nr. of iterations. Trying to check why.

mdymczyk:
@pseudotensor I understand but for now, unless someone can point me to a different solution (haven't found/figured out any myself) it's either this or we leave it as it was an tell the users that for certain datasets if you use floats you will see a large fluctuation in results and running time. If they want to be very precise they can run with doubles (which are, as you mentioned much slower - whole Homesite on k=1000 runs for me in ~3min on a 1080).

The only other solution I can come up with would be to sort the whole data matrix by key (label) and then do a reduction, like `thrust::reduce`. I am not sure, though 1) whether this would actually solve the problem 2) how slow would the sorting and reduction be 3) all the APIs I found for such methods are struggling with our data representation (matrix as a vector).

pseudotensor:
Ya, I'm just saying let people use float32 if they want or use float64 if want, but I wouldn't force float32 to use float64 internally if it slows it down alot.  And I wouldn't use a different technique with a much larger footprint, because probably someone using float32 is worried about both speed and footprint.

Also depends upon what you mean by very unstable in time and results.  How are you quantifying this?  Perhaps you are very strict, but in many applications you may not need to be.

mdymczyk:
@pseudotensor on the whole Homesite dataset, with exactly same parameters (n_gpus, random_state, k, max_iterations) for k=1000 I see it run for anywhere between 30 and 150 iterations on 1 GPU and anywhere between 30 and 300 iterations for 2 GPUs. So sometimes 2 GPUs are almost 2x as fast, sometimes they are 2x as slow.

When I compare the predicted labels for different runs they differ by 0-5%.

All depends on the order in which the summation kernels are being run by CUDA. If the order is different and we are working on numbers which overflow floating point operations, which generates rounding errors, it will produce different results.

mdymczyk:
On another note, I'm really surprised this hasn't been noticed while doing the initial benchmark several months ago? The code which causes the instability hasn't changed since then.

mdymczyk:
Another way to go about this would be to add a flag (either a parameter or env variable), which would turn on this feature (using doubles for centroid accumulation) for people who want to run floats with higher precision.

pseudotensor:
Ya, if code same and not noticed before, odd.  Do you have records of the original benchmarks, like in plots or otherwise?

mdymczyk:
@pseudotensor no but will check out an old version (from August) and try to rerun and see.

mdymczyk:
@pseudotensor I checked out 6e2800b01f73e5f2fc4e8e187b8f6fbe11910f08 (`Tue Jul 25`) - this was the last commit before Navdeep and I started contributing.

I see the same problem happening (actually even worse) - very unstable running times and results:

screen shot 2017-12-15 at 9 45 40 am

The difference in the runtime compared to the benchmark is probably because I'm running with a higher threshold (stops earlier).

mdymczyk:
So how about, for now, I just add a flag (but not sure whether it should be a parameter or an env variable?) which will by default be `false` but if turned on I will use `double` matrix as the accumulator for summation? This way the user can choose (and it will still be much faster than running the whole thing on `double`).

pseudotensor:
Sounds good.

mdymczyk:
Actually, my bad! Completely forgot that the old version was, by default, randomly shuffling the data. When I ran it with `selectstrat` which doesn't do any shuffling I'm getting very reproducible results! Which brings me back to square one... Did a git diff on the piece of code which is not deterministic currently and it's exactly the same as from that version.

pseudotensor:
It's still progress.  Keep looking, I'm sure you'll figure it out.

mdymczyk:
After debugging much more I noticed, that in the end the summation method for floats does introduce certain level of error depending on the order they are called but its negligible! The main reason, which I was able to only figure out after printing out all the parameters passed to that method, was because indeed one of the vectors differed each run!

After adding synchronizations here and there the single GPU variant is working fine now.

The multi GPU variant still fluctuates so I need to look into it.

I'm flabbergasted why using `double` sized accumulators fixed this issue in my previous solution but at least now I'm on track to finally fix this.

pseudotensor:
Ya, flabbergasted is good sign.  Means you are closer to understanding.  A blank mind would be worse.

mdymczyk:
Closing temporarily since this isn't a bug per se, needs a bit more discussion how should be handled.

PR #364, opened 2017-12-14
Initial work for R package with working KMeans example

terrytangyuan:
This is an initial PR for the work-in-progress R package is in r-lib branch. Check out this working example of using fit(), transform(), score(), predict(): https://github.com/h2oai/h2o4gpu/blob/r-lib/src/interface_r/tests/testthat/test_model.R. 

cc: @ledell @navdeep-G @mdymczyk @pseudotensor


pseudotensor:
Hi Terry, nice work!  I'll let @navdeep-G and @ledell comment on the R part.

My only comment would be the test you added for kmeans needs to be part of what is tested during "dotest" in the Makefile .  At least if the kmeans fails to run it should fail the makefile call so that jenkins knows something is wrong.  I'm not sure how you want to test it, but probably just adding another line to the "dotest:" in Makefile and running the script should be enough maybe -- just a pass/fail.

You can also add accuracy checks inside your script to fail if not accurate enough.

Thanks!

src/interface_r/DESCRIPTION
```diff
@@ -0,0 +1,28 @@
+Package: h2o4gpu
+Type: Package
+Title: h2o4gpu in R
+Version: 0.0.1.9999
+Authors@R: c(
+  person("Yuan", "Tang", role = c("aut", "cre"), email = "terrytangyuan@gmail.com"),
+  person("H2O.AI", role = c("cph", "fnd")))
```
navdeep-G: For this can you make the authors just `H2O.ai`.

src/interface_r/DESCRIPTION
```diff
@@ -0,0 +1,28 @@
...
+Maintainer: Yuan Tang <terrytangyuan@gmail.com>
```
navdeep-G: For the maintainer can you make it multiple people? I forget if you can. Would be good if its me, you, and @ledell.

src/interface_r/R/install.R
```diff
@@ -0,0 +1,16 @@
+#' Install h2o4gpu from a path to wheel file
+#' @export
+install_h2o4gpu <- function(wheel_path) {
+  # TODO: Support other install methods
```
navdeep-G: Other install methods would be virtualenv, which is what most devs for this project use.

src/interface_r/R/install.R
```diff
@@ -0,0 +1,16 @@
...
+  conda_env <- "r-h2o4gpu"
+  # Remove this once PR is merged to reticulate
+  conda_path <- "/home/terry/miniconda3/bin/conda"
```
navdeep-G: Hard coded path.

src/interface_r/R/package.R
```diff
@@ -0,0 +1,43 @@
...
+    environment = "r-h2o4gpu",
+    
+    on_load = function() {
+      check_compatibility(displayed_warning)
+      search_env <- "r-h2o4gpu"
+      conda_path <- "/home/terry/miniconda3/bin/conda"
+      if (search_env %in% reticulate::conda_list(conda = conda_path)$name) {
+        use_condaenv(search_env, conda = conda_path)
+      } else {
+        message("h2o4gpu Python package is not installed. Please use install_h2o4gpu(path_to_wheel) to install.")
```
navdeep-G: There is a s3 path to a .whl file. Shown in the h2o4gpu README

src/interface_r/R/package.R
```diff
@@ -0,0 +1,43 @@
...
+  delay_load <- list(
+    
+    priority = 5,
+    
+    environment = "r-h2o4gpu",
+    
+    on_load = function() {
+      check_compatibility(displayed_warning)
+      search_env <- "r-h2o4gpu"
+      conda_path <- "/home/terry/miniconda3/bin/conda"
```
navdeep-G: Hard coded path. Need to account for virtualenv, as well.

src/interface_r/R/type_resolvers.R
```diff
@@ -0,0 +1,18 @@
...
+  if (is.null(x))
+    x
+  else
+    as.integer(x)
+}
+
+resolve_model_input <- function(x) {
+  if (is.matrix(x)) {
+    x
+  } else if (is.data.frame(x)) {
```
navdeep-G: This should cover `data.table`, as well. But good to double check.

src/interface_r/R/install.R
terrytangyuan: This will involve a little more work. I'll work on those in future PRs

src/interface_r/R/install.R
terrytangyuan: Good catch!

src/interface_r/R/package.R
terrytangyuan: Yes, this is a method that user can call and they can specify a path, include s3 path. I'll add more docs in future PRs.

src/interface_r/DESCRIPTION
navdeep-G: Or you can add me, @ledell and yourself and give all "aut" and "cre" in addition to H2O.ai

navdeep-G:
```

Comment (navdeep-G) [reply src/interface_r/DESCRIPTION]: I would like this: 
```
person("Yuan", "Tang", role = c("aut", "cre"), email = "terrytangyuan@gmail.com"),
person("Navdeep", "Gill", role = c("aut", "cre"), email = "navdeep@h2o.ai"),
person("Erin", "LeDell", role = c("aut", "cre"), email = "erin@h2o.ai"),
person("H2O.AI", role = c("cph", "fnd")))
```

Comment (terrytangyuan) [reply src/interface_r/DESCRIPTION]: Sounds good. Will do.

File: src/interface_r/DESCRIPTION [hunk 8]
```diff
@@ -0,0 +1,29 @@
+Package: h2o4gpu
+Type: Package
+Title: h2o4gpu in R
+Version: 0.0.1.9999
```
ledell: Let's use [Hadley's recommendation](http://r-pkgs.had.co.nz/description.html) of starting the package at 0.0.0.9000 and then bump the last digit every time someone adds something new (e.g. 0.0.0.9001, 0.0.0.9002, etc).  When all the algos are added, then we can take this out of development mode.

terrytangyuan: Sounds good. Will change.

src/interface_r/R/install.R
```diff
@@ -0,0 +1,14 @@
...
+  conda_env <- "r-h2o4gpu"
+  if (!conda_env %in% reticulate::conda_list()$name) {
+    conda_create(envname = conda_env)
+  } else {
+    conda_remove(envname = conda_env)
+    conda_create(envname = conda_env)
+  }
+  use_condaenv(conda_env)
+  conda_install(envname = conda_env, packages = wheel_path, pip = TRUE, pip_ignore_installed = TRUE)
+}
```
navdeep-G: So the idea here is to create a conda_env `r-h2o4gpu` no matter what?

terrytangyuan: Or a virtualenv (future work) named `r-h2o4gpu` (unfortunately this has to be a hard-coded string). In TensorFlow R binding, we had something similar `r-tensorflow`. This way R users don't need to activate the environments themselves. We can provide this as R `options` too so if any users want to change this, they can do so as well.

terrytangyuan: e.g. 
```
options(h2o4gpu_env_name='r-h2o4gpu-2')
options(h2o4gpu_env_load_method='virtualenv')
```

navdeep-G: Gotcha and why is the `virtualenv` future work? Is it not available within reticulate?

terrytangyuan: It's not fully available yet. There are `use_conda`, `conda_create`, `conda_list`, etc. but there's only `use_virtualenv` at this moment.

navdeep-G: Ah, I see. I think we can expose that for folks who know how to build a virtualenv. What do you think of that?

src/interface_r/R/kmeans.R
```diff
@@ -0,0 +1,39 @@
...
+  random_state = NULL,
+  copy_x = TRUE,
+  n_jobs = 1,
+  algorithm = 'auto',
+  gpu_id = 0,
+  n_gpus = -1,
+  do_checks = 1,
+  backend = 'auto'
+) {
+  model <- h2o4gpu$KMeans(
```
navdeep-G: So the user will pass in data via `%>% fit(x)`? Is this the behavior we want? I would think we want to pass in `X` directly with the rest of the params much like other R functions.

src/interface_r/R/type_resolvers.R
ledell: This is okay; `is.data.frame(x)` returns TRUE when passed a data.table or tibble as well.

src/interface_r/R/kmeans.R
terrytangyuan: We can but it's nicer to separate them conceptually, e.g. initialize the model object and then decide what to do with it (fit, predict, transform, etc.). One use case is to load a pre-trained model and transform new data or continue training using new data (you only initialize the model once and then use it for multiple purposes, e.g. transform and fit). It's more consistent with sklearn, keras, etc.

ledell:
@terrytangyuan Can you comment on which of these R files were auto-generated by Reticulate and which were manually created?

src/interface_r/R/kmeans.R
terrytangyuan: Note that you can also do `fit(model, x)`.

terrytangyuan:
@ledell `kmeans.R` was generated by `reticulate::py_function_wrapper()` in additiona with some manual edits to make sure parameters are passed in with the appropriate types, e.g. `as_integer()` so that R users don't need to do `1L` (this is unavoidable due to the nature of reticulate) but it's only one time effort.

src/interface_r/R/install.R
terrytangyuan: You mean exposing those options?

terrytangyuan:
@ledell What I meant was that almost everything was coded manually so we can provide appropriate checks for each argument. Given that we don't have too many algorithms yet we could probably just have this one time efforts and do everything manually. I doubt that we will spend too much time on maintaining since they are basically thin wrappers of Python API (only need to update when we have discussion in #api-updates.  What do you think?

terrytangyuan:
I decided to remove conda related stuff and let users manually do this instead. 

So the new workflow for users would be like the following:
```
library(h2o4gpu)
use_conda("xxxx") # or use_virtualenv("xxx") or activate the choice of environment or activate the environment before executing this R script
model <- KMeans(...)
```
Are you guys okay with this? @ledell @navdeep-G If there are other things that you'd like me to change in this PR, please let me know.

navdeep-G:
@terrytangyuan I think removing the conda stuff and letting the user set up their environment is fine. We can revisit a better solution in the near future, but this won't hinder anyones experience imho.

src/interface_r/interface_r.Rproj
```diff
@@ -0,0 +1,17 @@
+Version: 1.0
```
navdeep-G: Is this file needed?

terrytangyuan: Yes it's better to have a Rproj so you can do a lot of stuff in RStudio more easily. Many R projects have this included in the repo.

src/interface_r/R/kmeans.R
navdeep-G: Sounds good to me!

src/interface_r/interface_r.Rproj
navdeep-G: Ah, yes, got this confused with the other .Rproj files. Not all are needed, of course.

src/interface_r/tests/testthat/test_model.R
```diff
@@ -0,0 +1,24 @@
...
+
+test_that("Models accept different types of inputs", {
+  raw_input <- list(list(1, 1), list(1, 4), list(1, 0))
+  x <- np$array(list(list(1, 1), list(1, 4), list(1, 0)))
+  
+  # Test support for data.frame
+  model <- h2o4gpu.kmeans(n_clusters = 2L, random_state = 1234L) %>% fit(as.data.frame(x))
+
+  # Test support for list
+  model <- h2o4gpu.kmeans(n_clusters = 2L, random_state = 1234L) %>% fit(raw_input)
```
navdeep-G: I would also check `fit(model, X)`, `predict(model, X)`, `transform(model, X)`, and `score(model, X)`

terrytangyuan: Done!

PR #367, opened 2017-12-19
run tests on ec2 P32xlarge

mstensmo:
Hi, here's what I think it should look like to get a whl from s3 and then run tests on it. Once this file is correct I'll also make files for the other settings, or maybe we can just do all of them in the same test.

Open questions. 
1) will it load the utils when in a subdir?
2) Can I duplicate this block and then run all tests as stages?


pseudotensor:
Cleaned-up and worked on this with @mstensmo separately, already in master.

PR #368, opened 2017-12-20
Cleanup jenkins files to avoid duplicate code

pseudotensor:


pseudotensor:
@abal5 How do I do this?  :)

pseudotensor:
https://stackoverflow.com/questions/9136328/including-a-groovy-script-in-another-groovy

https://stackoverflow.com/questions/41739468/groovy-reports-that-a-file-doesnt-exists-when-it-really-is-present-in-the-syste

PR #382, opened 2018-01-09
[R] Support character/factor columns for labels

terrytangyuan:
They are commonly used types and we should convert them to numeric so h2o4gpu knows how to handle them. Also give warning to users so they know what to expect.


navdeep-G:
LGTM. Thoughts @ledell ?

terrytangyuan:
@ledell I merged this for now so it won't block my other work but feel free to still comment and I'll address them in new PRs.

PR #383, opened 2018-01-09
[R] Use pyenv binary for testing only when it exists

terrytangyuan:


terrytangyuan:
@navdeep-G @ledell This is just a very simple change. I just want to get more eyes on it. Also, do simple changes like these need to go through PR going forward? 
(I won't merge this yet until master branch is fixed)

navdeep-G:
@terrytangyuan I believe all changes should go through PR. No matter how minor. Just a simple rule to follow.

PR #385, opened 2018-01-11
[R] Support regression and added random_forest_regressor as an example

terrytangyuan:
* Support regressors
* Added the handling of regressor and classifier's target column separately
* Added `random_forest_regressor` wrapper and its unit test.

Note that the wrapper functions, e.g. `random_forest_regressor`, will be generated automatically in the future.


ledell:
@terrytangyuan Can you add a brief description of the updates?

terrytangyuan:
@ledell Done. Let me know if anything needs to be clarified better.

PR #397, opened 2018-01-20
Add missing imports for tsvd and update default family for scikit fallback

navdeep-G:
* Add missing imports to `__init__ ` files
* Add `gpu_id` to `TruncatedSVD` class (previously added to `TruncatedSVDH2O` class)
* Update test to account for default family `cusolver`, which switches to `arpack` if `do_sklearn` is set to `true` with algorithm set to `cusolver`.


pseudotensor:
on any imports they need to only be ised where needed. localize all inports instead of global, else will break DAI

pseudotensor:
those explicit inports may be fine, but ibside thise files any imports should be localized unless really always used like numpy

navdeep-G:
These imports are just the other class files we have besides the wrapper. This also helps expose it for the R-api, which had to be done earlier today.

PR #398, opened 2018-01-21
Fixes #400: More directly write wrappers to a file instead of relying on sink()

### LINKED ISSUE #400: Autogeneation of R model wrappers may not work on certain environment
Autogeneration of R model wrappers added in https://github.com/h2oai/h2o4gpu/pull/388 doesn’t work sometimes for users on server with issues with sink(). The generated file is blank while individual wrappers can be generated successfully. We need to switch to use more direct write instead of relying on sink(). Can be reproduced using this script: https://github.com/h2oai/h2o4gpu/blob/12eaa761e2c5f145fb05ef59d1a401c6ddcbf125/src/interface_r/scripts/gen_wrappers.R

terrytangyuan:
@navdeep-G This should fix the problem you are seeing. Could you give it a try to see if it's fixed?


mdymczyk:
@terrytangyuan for the future could we have a github issue for every PR, with a description of the problem and how to reproduce it? Just in case someone goes on holidays/becomes available and someone else needs to review it.

terrytangyuan:
Sure, I submitted an issue here https://github.com/h2oai/h2o4gpu/issues/400! I don’t think we need issue for every PR though since it seems to have too much overhead...for big ones I do agree. We can certainly use the PR descriptions to briefly clarify the problem for small PRs.

PR #406, opened 2018-01-23
Monika pydaal

stjoern:
Linear Regression from Intel DAAL included to the Solvers.


navdeep-G:
Many commits in here. Might need to clean up a bit.

PR #408, opened 2018-01-23
Monika pydaal new2

stjoern:
IntelDaal included to the project only for calculating Linear Regression.

+ I have created prebuild of IntelDaal only for Linux-x86_64 https://s3.amazonaws.com/intel-daal/daal-linux_x86_64.tar.gz
+ For testing, please run in virtual env. `h2o4gpu/scripts/daal/daal_preinstalled_packages.sh`
+ The script will set LD_LIBRARY_PATH locally

@mdymczyk : #406 and #407 can be closed and reviewed only #408.
@mdymczyk : I am working on points 1,2,3. Thank you.


mdymczyk:
Can we close #406 and #407 and just review this one? I'm assuming they are the same thing?

mdymczyk:
Had a quick look and so far looks good. Things we'll need:

1) benchmarks against other CPU implementations, SKLearn would be a good start. 
2) If possible (not sure if we have any at the office, maybe AWS?) check the performance on nonDAAL processors (I'm assuming this will run on any CPU but slower, right?). So maybe some AMDs? Maybe ask @CreatureDev (Patrick Rice) about what he's working with.
3) tests with assertions - how is the performance/accuracy here compared to SKLearn or other frameworks? Are we getting more or less the same results on the same datasets? if not then why? Something like this for starters https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/linear_model/tests/test_base.py

pseudotensor:
The latest is still complaining about some daal stuff:
```

>       self.model_daal = DaalRegression(fit_intercept=fit_intercept,

                                               normalize=normalize)

E       NameError: name 'DaalRegression' is not defined


```

pseudotensor:
I fixed the conflict in dockerfile-build.

pseudotensor:
Hi Monika.  Can you resolve the Makefile conflicts?  We should be able to still compile for cpu without intel daal, so please keep both options.  Thanks!

mdymczyk:
Hey @stjoern looking good! As Jon mentioned, can we resolve the Makefile conflicts with master and squash this PR into one commit? I'll do a proper review after that but so far looks ok. Also can you paste some benchmarks (think you mentioned you did them already) here? Thanks!

mdymczyk:
@stjoern went through the code, all looks good! 2 more things from my side and I think we can merge if @pseudotensor and @navdeep-G don't have any objections:

1) need some tests with actual assertions - I think it would make sense to run the same test data (can be something basic) with SKLearn, 4gpu and daal backends and check if we are getting similar results. You can ask @khevn for ideas, he can come up with some basic scenarios for you, he's good at that.

2) add a benchmark graph here and somewhere in the repo (maybe a new folder with benchmarks?) - saying on what data it was ran and using which code version.

stjoern:
I am currently working on SVD, where I am preparing jupyter workbook, maybe I could add new jupyter to linear regression too; and added scaling (min-max, zscoring - default zscoring for Linear Regression).
@mdymczyk : I've added so far only one test, pls. see **test_fit_linear_regression_daal_vs_sklearn**

mdymczyk:
@stjoern yes jupyter notebook is a good start!

I saw the test. The issue I was addressing above is that is has no assertions (I think some of our other tests have/had same problems so we'll have to fix that). It only checks for run time but we also need to know whether our solvers also result in proper predictions (using `assert` etc). Something like they have for sklearn https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/linear_model/tests/test_base.py

Ideally at least for basic cases results for SKLearn, H2O4GPU and DAAL should be the same.

mdymczyk:
@stjoern the tests look good! We still need L1/L2 regularizations, right?

stjoern:
@mdymczyk we have already implemented L2 together with test cases.

mdymczyk:
@stjoern ok lets discuss tomorrow during standup how hard would it be to implement L1. If it's a major effort then we can make an issue for it and do it in the future. We'd have to handle L1 case should the user try to run L1 with DAAL backend appropriately - either throw an exception that it's not supported or log a warning and fallback to Scikit.

Also one thing I'm worried about is whether this DAAL enabled build would be generic enough - could you consult with Tom K on which platforms we need DAI and h2o4gpu to run? I know they are working on some obscure platforms and if we merge stuff which makes those fail we'll have problems :-)

PR #413, opened 2018-01-25
Fixes #410: Install necessary R deps for fullinstall

### LINKED ISSUE #410: fullinstall broken.  Needs R with specific version, etc.
I don't have it, but it should be installed as part of dependency.
```
 Test R package
/usr/bin/R-3.1.0 -e 'devtools::test("src/interface_r")'
/bin/bash: /usr/bin/R-3.1.0: No such file or directory
Makefile:468: recipe for target 'dotest' failed
make: *** [dotest] Error 127
jon@pseudotensor:~/h2o4gpu$ 
```

terrytangyuan:


navdeep-G:
PR lgtm. Maybe can check on a system that doesn't have R? cc @pseudotensor

terrytangyuan:
The Docker environment doesn’t have R so i think it’s already been tested there. Let me know if you encounter any new issues though.

pseudotensor:
Ya, issue was docker was installing R.  We need those instructions in the readme as part of requirements.  But definitely shouldn't point to specific R version 

Ya, I can test it.  I presume I need to follow some README docs, not just do make fullinstall that shouldn't download R.

pseudotensor:
I see, you made: scripts/install_r_deps.sh .  Just point to this script in the requirements when saying R is required.

terrytangyuan:
@pseudotensor Yes, that is the script. I just added the link to the script in DEVEL.md. Let me know if there's anything else.

pseudotensor:
And add something in README about requiring R.

The other question is... if R doesn't exist, do we want everything to still work and compile but just not have R?   I think so.  Is that tested?  Currently at least the fullinstall fails looking for R.

terrytangyuan:
Where specifically should I add the requirements? The requirements for R package is documented in its own directory: https://github.com/h2oai/h2o4gpu/tree/master/src/interface_r (this is the only place where R users would care about). I think README in root directory should only contain the general requirements. 

If R doesn't exist, that script I added will install R for you automatically. This is tested in Jenkins as part of the Dockerfile where R doesn't exist.  Feel free to give it a try.

pseudotensor:
Ok, make fullinstall doesn't fail for me.  I'll assume it handles R stuff in reasonable way and merge this PR once the last merge with master passes.

PR #430, opened 2018-02-01
Update R api to include only h2o4gpu classes instead of sklearn wrappers

navdeep-G:
* The main point of this PR is to update the R API to include classes that are mainly built by the h2o4gpu project, i.e., classes that do not wrap the sklearn API as this might not be pertinent to R users.
* This gives a benefit of not having to worry about various classes that do not pertain to the R ecosystem (it makes sense for Python as scikit is the go-to library for machine learning (for the most part)).
* This also simplifies the API a bit, which is always good.
* The number of solvers is now reduced to PCA, KMeans, SVD, ElasticNet(split into regressor and classifier), GBMRegressor, GBMClassifiier, RandomForestRegression, and RandomForestClassifier
* Update ElasticNet class to include a superset of scikit learn and h2o4gpu params
* Update n_iter in TruncatedSVD from 5 to 100
* Added `backend` param to R API which is set to `h2o4gpu` by default. User can set to `sklearn`


src/interface_r/R/auto_generated_wrappers.R
```diff
@@ -225,191 +225,57 @@ h2o4gpu.gradient_boosting_regressor <- function(
...
-	tol = 0.01,
-	solver = "auto",
-	random_state = NULL,
-	n_gpus = -1L,
-	glm_stop_early = TRUE,
-	glm_stop_early_error_fraction = 1.0,
-	verbose = FALSE,
-	backend = "auto") {
+	verbose = 0L,
+	family = "elasticnet",
```
ledell: What does `family = "elasticnet"` mean?

ledell:
@navdeep-G To follow the convention, shouldn't ElasticNet also have a classifier and regressor function like the other supervised algos do?  It seems like you are trying to infer the task (classification vs regression) by using the `family` argument, but we could have both, as long as regression supports multiple familes (gaussian, poisson, etc), but I don't think it does support more families, does it?

src/interface_r/R/auto_generated_wrappers.R
navdeep-G: Regression

navdeep-G:
@ledell Yes, we can do this but the `LogisiticRegression` function follows scikit params (so not what we have have in the ElasticNet class fully), which is not ideal I think since we are missing out on the full capability it has to offer.

navdeep-G:
Think of ElasticNet as a class that has the full capability of h2o4gpu while the LogisticRegression class follows what scikit has only.

navdeep-G:
And no, other families are not supported.

navdeep-G:
What I can do is make an ElasticNetClassifier & ElasticNetRegressor like you said where family is set for each. This will have to be done in Python first.

ledell:
@navdeep-G I see you've updated `gen_wrappers.R`, so I just wanted to check that you only made manual edits to this file and then auto-generated the R wrappers from here?

navdeep-G:
@ledell Yes, that’s correct.

src/interface_r/R/auto_generated_wrappers.R
ledell: We talked about this offline, it's the backend ElasticNet function's way of distinguishing between classification and regression.  "elasticnet" means regression and "logistic" means regression (this does not many much sense so we should change this).  More discussion [here](https://github.com/h2oai/h2o4gpu/issues/431).

src/interface_r/scripts/gen_wrappers.R
```diff
@@ -49,51 +49,51 @@ write_wrapper("h2o4gpu$GradientBoostingRegressor",
...
+#             r_function = "h2o4gpu.lasso_regressor",
+#             nullable_int_params = common_nullabl_int_params,
+#             class_tags = 'c("regressor")',
+#             file_name = file_name)
+# 
+# write_wrapper("h2o4gpu$Ridge",
+#             r_function = "h2o4gpu.ridge_regressor",
+#             nullable_int_params = common_nullabl_int_params,
+#             class_tags = 'c("regressor")',
+#             file_name = file_name)
```
terrytangyuan: Are you planning to support these going forward in H2O wrappers? If yes, could you put a comment here? If no, please simply remove them for now.

terrytangyuan:
Looked at your conversation. I agree with you guys here - it would be better to make separate wrappers for classification and regression so it's clearer and more consistent to the user.

src/interface_r/scripts/gen_wrappers.R
navdeep-G: I am not sure yet so I will add a note.

terrytangyuan: Another question is: if sklearn is a requirement to build the package, should we simply keep it as it is since it's already installed and we can still keep the wide coverage of types of models.

navdeep-G: I wouldn't think that's not ideal since the sklearn requirement is needed for a specific reason: to be compatible with current scikit like pipelines in Python. Doesn't really apply in this scenario.

navdeep-G: A good rule of thumb is if the class has a `backend` argument, then it is not needed for the R api as we wouldn't expect an R user to want to use scikit.

terrytangyuan: okay I see. Thanks for the clarification. Ping us when this is ready for review.

navdeep-G:
@terrytangyuan @ledell Made changes we discussed here: 3e9f5cd

navdeep-G:
One more fix needed. Please wait. @terrytangyuan @ledell

src/interface_r/R/auto_generated_wrappers.R
```diff
@@ -219,199 +215,119 @@ h2o4gpu.gradient_boosting_regressor <- function(
...
-	n_gpus = -1L,
+	tol_seek_factor = 0.1,
 	lambda_stop_early = TRUE,
 	glm_stop_early = TRUE,
 	glm_stop_early_error_fraction = 1.0,
-	verbose = FALSE,
-	backend = "auto") {
+	max_iter = 5000L,
+	verbose = 0L,
+	 store_full_path = 0L,
```
terrytangyuan: I wonder where this additional space comes from but no big deal

src/interface_r/scripts/gen_wrapper_utils.R
```diff
@@ -46,7 +48,17 @@ gen_wrapper <- function(
             param <- paste0("as_nullable_integer(", param, ")")
           }
           suffix <- ifelse(i < length(params), ",", "\n  )")
-          write_line(paste0("    ", params[[i]], " = ", param, suffix))
+          if(param == "backend"){
+            write_line(paste0("    ", params[[i]], " = ", '"h2o4gpu"', suffix))
+          } else if (param == "family"){
+            if(class_tags == "classifier"){
```
terrytangyuan: actually this should be `grepl("classifier", class_tags)`

navdeep-G:
All fixes are in @ledell @terrytangyuan

terrytangyuan:
@navdeep-G Looks good. I preferred your fix that's more specific for this particular estimator.

src/interface_r/R/auto_generated_wrappers.R
ledell: I think it's caused by the fact that there is no space after the Python [docstring name](https://github.com/h2oai/h2o4gpu/blob/c0d55138e5ca0a683bccae8f06f5757bd22e0a2a/src/interface_py/h2o4gpu/solvers/elastic_net.py#L75).  

This one looks like `:param int store_full_path: Whether to store full solution for all alphas` 
where the others look like: `:param int max_iter : Maximum number of iterations. Default is 5000`

That's the only difference I can notice in Python.

src/interface_r/R/model.R
```diff
@@ -12,13 +12,21 @@ attach_attrs_to_model <- function(r_model_obj) {
   attrs_exclude_from_attach <- c(
     "fit", "fit_predict", "fit_transform", "score", "predict", "transform",
     "init")
-  model_attrs <- names(r_model_obj$model$model)
+  if(grepl("H2O", as.character(r_model_obj$model))){
```
ledell: Nitpick: Space missing after `if(`.  There are probably more code formatting issues scattered throughout the code, so we can ignore for now if you want and apply [lintr](https://github.com/jimhester/lintr) at some point before release to clean up spacing.

navdeep-G: I think a lintr is the right way to go. I can do that in a sep PR

src/interface_r/R/auto_generated_wrappers.R
```diff
@@ -219,199 +215,119 @@ h2o4gpu.gradient_boosting_regressor <- function(
...
+    tol_seek_factor = tol_seek_factor,
     lambda_stop_early = lambda_stop_early,
     glm_stop_early = glm_stop_early,
     glm_stop_early_error_fraction = glm_stop_early_error_fraction,
-    verbose = verbose,
-    backend = backend
+    max_iter = as.integer(max_iter),
+    verbose = as.integer(verbose),
+    family = "logistic",
+    store_full_path = as.integer(store_full_path),
```
ledell: Same comment as elsewhere, but I think `store_full_path` should be changed to a boolean in Python (and then here).

src/interface_r/R/auto_generated_wrappers.R
```diff
@@ -219,199 +215,119 @@ h2o4gpu.gradient_boosting_regressor <- function(
...
-    n_gpus = as.integer(n_gpus),
+    tol_seek_factor = tol_seek_factor,
     lambda_stop_early = lambda_stop_early,
     glm_stop_early = glm_stop_early,
     glm_stop_early_error_fraction = glm_stop_early_error_fraction,
-    verbose = verbose,
-    backend = backend
+    max_iter = as.integer(max_iter),
+    verbose = as.integer(verbose),
+    family = "logistic",
```
ledell: Let's change `family` in Python to "gaussian"/"binomial" in the general ElasticNet class.  Nothing to do here in this PR though.

src/interface_r/R/auto_generated_wrappers.R
```diff
@@ -219,199 +215,119 @@ h2o4gpu.gradient_boosting_regressor <- function(
...
-    solver = solver,
-    random_state = as_nullable_integer(random_state),
-    n_gpus = as.integer(n_gpus),
-    glm_stop_early = glm_stop_early,
-    glm_stop_early_error_fraction = glm_stop_early_error_fraction,
-    verbose = verbose,
-    backend = backend
+    verbose = as.integer(verbose),
+    family = "elasticnet",
+    store_full_path = as.integer(store_full_path),
```
ledell: Same comment as elsewhere, this should be changed to boolean in Python (then here).

src/interface_r/R/auto_generated_wrappers.R
```diff
@@ -219,199 +215,119 @@ h2o4gpu.gradient_boosting_regressor <- function(
...
+	alphas = NULL,
+	lambdas = NULL,
+	double_precision = NULL,
+	order = NULL) {
 
-  model <- h2o4gpu$ElasticNet(
-    alpha = alpha,
-    l1_ratio = l1_ratio,
+  model <- h2o4gpu$ElasticNetH2O(
+    n_threads = n_threads,
```
ledell: @terrytangyuan This should be `as_nullable_integer(n_threads)` so we need to change something in the `gen_wrappers.R` function.  Seems like we could change [this line](https://github.com/h2oai/h2o4gpu/blob/master/src/interface_r/scripts/gen_wrappers.R#L19):

```
common_nullabl_int_params <- c("random_state")
```
to 
```
common_nullabl_int_params <- c("random_state", "n_threads")
```
By the way, what does `n_threads = NULL` mean?  Does it mean that we will use all threads available on the machine?  In H2O, we have the default `nthreads = -1` which means use nthreads == n cores on the machine.

src/interface_r/R/auto_generated_wrappers.R
```diff
@@ -219,199 +215,119 @@ h2o4gpu.gradient_boosting_regressor <- function(
...
-    class_weight = class_weight,
-    random_state = as_nullable_integer(random_state),
-    solver = solver,
-    max_iter = as.integer(max_iter),
-    multi_class = multi_class,
-    verbose = as.integer(verbose),
-    warm_start = warm_start,
-    n_jobs = as.integer(n_jobs),
+  model <- h2o4gpu$ElasticNetH2O(
+    n_threads = n_threads,
```
ledell: Should be `as_nullable_integer(n_threads)`.  See other comment about this on how to fix.

src/interface_r/R/auto_generated_wrappers.R
ledell: @terrytangyuan @navdeep-G On the other hand, should we even have an `n_threads` parameter in ElasticNet?  I guess that's only available on ElasticNet, but not the other algos?  Do you know if XGBoost will use multiple threads by default?  The other algos do not have `n_threads`, they just have `n_gpus`.  

And I guess if want to keep `n_threads` here, since the other algos don't have it, then it might be a problem to add it to `common_nullabl_int_params` in `gen_wrappers.R` and we'll have to explicitly make this a nullable int.

src/interface_r/R/auto_generated_wrappers.R
```diff
@@ -219,199 +215,119 @@ h2o4gpu.gradient_boosting_regressor <- function(
...
+    gpu_id = as.integer(gpu_id),
```
ledell: Do we know why only some of the algos have a `gpu_id`?  I see it in Elastic Net, Truncated SVD and K-Means, which I guess just means that only a subset of the tools we use underneath support this level of precision w/r/t computation.

src/interface_r/R/auto_generated_wrappers.R
```diff
@@ -446,57 +361,42 @@ h2o4gpu.kmeans <- function(
...
 	n_components = 2L,
-	algorithm = "cusolver",
-	n_iter = 5L,
+	algorithm = "power",
+	n_iter = 100L,
 	random_state = NULL,
 	tol = 1e-05,
-	verbose = FALSE,
-	backend = "auto",
+	verbose = 0L,
```
ledell: Why did this get changed from a boolean to an int?  I see that it's different when we use "TruncatedSVDH2O" versus "TruncasedSVD" but why do the two methods have different object types?

src/interface_r/R/auto_generated_wrappers.R
```diff
@@ -446,57 +361,42 @@ h2o4gpu.kmeans <- function(
...
   h2o4gpu_model(model, NULL)
 }
 
 #' @export
 h2o4gpu.truncated_svd <- function(
 	n_components = 2L,
-	algorithm = "cusolver",
-	n_iter = 5L,
+	algorithm = "power",
+	n_iter = 100L,
```
ledell: Why does "TruncatedSVDH2O" have different default vaules for `algorithm` and `n_iter` as compared to the "TruncatedSVD" class?  This is not good because if you use this methods in R and Python, they will produce different results by default.

src/interface_r/R/auto_generated_wrappers.R
```diff
@@ -446,57 +361,42 @@ h2o4gpu.kmeans <- function(
...
+  model <- h2o4gpu$TruncatedSVDH2O(
     n_components = as.integer(n_components),
     algorithm = algorithm,
     n_iter = as.integer(n_iter),
     random_state = as_nullable_integer(random_state),
     tol = tol,
-    verbose = verbose,
-    backend = backend,
+    verbose = as.integer(verbose),
+    n_gpus = as.integer(n_gpus),
```
ledell: So it seems like "TruncatedSVDH2O" has the ability to specify `n_gpus` but "TruncatedSVD" doesn't?  That seems odd considering both of them have `gpu_id`.

src/interface_r/R/auto_generated_wrappers.R
```diff
@@ -446,57 +361,42 @@ h2o4gpu.kmeans <- function(
...
     n_components = as.integer(n_components),
-    copy = copy,
-    whiten = whiten,
-    svd_solver = svd_solver,
-    tol = tol,
-    iterated_power = iterated_power,
-    random_state = as_nullable_integer(random_state),
-    verbose = verbose,
-    backend = backend
+    whiten = whiten
```
ledell: Hmm, making these switches to the "-H2O" classes is changing a bunch of parameters, so R and Python H2O4GPU will offer different versions of the algorithms.  Let's discuss this further, I think.

src/interface_r/R/auto_generated_wrappers.R
```diff
@@ -219,199 +215,119 @@ h2o4gpu.gradient_boosting_regressor <- function(
...
+	gpu_id = 0L,
+	n_gpus = -1L,
 	fit_intercept = TRUE,
-	normalize = FALSE,
-	precompute = FALSE,
-	max_iter = 5000L,
-	copy_X = TRUE,
+	lambda_min_ratio = 1e-07,
+	n_lambdas = 100L,
+	n_folds = 5L,
```
ledell: Do any other algorithms do CV under the hood?  Is this a property that we get for free with POGS?  I'm not sure that we want to mix and match validation strategies, meaning, allow internal CV in some algorithms but not in others.  That seems like it could create issues with validation & understanding what data has been used to select internal parameters.

src/interface_r/R/auto_generated_wrappers.R
```diff
@@ -219,199 +215,119 @@ h2o4gpu.gradient_boosting_regressor <- function(
...
+h2o4gpu.elastic_net_classifier <- function(
+	n_threads = NULL,
+	gpu_id = 0L,
+	n_gpus = -1L,
 	fit_intercept = TRUE,
-	normalize = FALSE,
-	precompute = FALSE,
-	max_iter = 5000L,
-	copy_X = TRUE,
+	lambda_min_ratio = 1e-07,
```
ledell: In the sklearn wrapper, is there a `lambda_min_ratio` equivalent?  What are we setting this to on the sklearn side before passing it to ElasticNetH2O?  (I assume it's just hardcoded to 1e-07).

ledell: What happened to `normalize`, does this mean that we're no longer normalizing, or normalizing by default?  I guess sklearn does this outside of ElasticNetH2O, so maybe we don't have the ability to do that within ElasticNetH2O.  That seems like a loss of good functionality... why not just utilize sklearn's ability to normalize.  Another reason that we may consider switching back t using the sklearn wrappers instead of the H2O wrappers directly... :thinking:

src/interface_r/R/auto_generated_wrappers.R
```diff
@@ -219,199 +215,119 @@ h2o4gpu.gradient_boosting_regressor <- function(
...
+    verbose = as.integer(verbose),
+    family = "elasticnet",
+    store_full_path = as.integer(store_full_path),
+    lambda_max = lambda_max,
+    alpha_max = alpha_max,
+    alpha_min = alpha_min,
+    alphas = alphas,
+    lambdas = lambdas,
+    double_precision = double_precision,
+    order = order
```
ledell: Looks like we are getting all these new parameters now that we're no longer using the sklearn wrappers.  Not sure how to document all of this... basically we have two completely different algorithms (not just interfaces) between R and Python now.

src/interface_r/R/auto_generated_wrappers.R
navdeep-G: I have updated this

navdeep-G:
@ledell I have addressed most of your comments (I think). Please review.

src/interface_py/h2o4gpu/solvers/elastic_net.py
```diff
@@ -1382,22 +1397,26 @@ def __init__(
...
-        lambdas = None
+        # n_threads = None
+        # n_alphas = 1
+        # n_lambdas = 1
+        # n_folds = 1
+        # lambda_max = alpha
+        # lambda_min_ratio = 1.0
+        # store_full_path = 1
+        # alphas = None
+        # lambdas = None
```
terrytangyuan: why are these commented out?

navdeep-G: They will now be passed from the constructor. Will remove.

src/interface_r/scripts/gen_wrapper_utils.R
```diff
@@ -46,7 +48,18 @@ gen_wrapper <- function(
             param <- paste0("as_nullable_integer(", param, ")")
           }
           suffix <- ifelse(i < length(params), ",", "\n  )")
-          write_line(paste0("    ", params[[i]], " = ", param, suffix))
+          # if(param == "backend"){
+          #   write_line(paste0("    ", params[[i]], " = ", '"h2o4gpu"', suffix))
+          # } else
```
terrytangyuan: Are these safe to be removed as well?

navdeep-G: Yes, we want to expose the `backend` param to users.

navdeep-G:
@ledell I will be adding one more change to this. Will let you know when.

src/interface_py/h2o4gpu/solvers/elastic_net.py
```diff
@@ -1304,26 +1304,41 @@ class ElasticNet(object):
...
-            normalize=False,
-            precompute=False,
-            max_iter=5000,  #h2o4gpu
-            copy_X=True,
-            tol=1e-2,  #h2o4gpu
-            warm_start=False,
-            positive=False,
-            random_state=None,
-            selection='cyclic',
+            alpha=1.0, #scikit
```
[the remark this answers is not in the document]
navdeep-G: Alpha and l1_ratio are scikit specific only I think.

PR #441, opened 2018-02-07
Minor fix to tsvd verbose output

navdeep-G:


mdymczyk:
Any reason we are not using the logger class? (e.g. https://github.com/h2oai/h2o4gpu/blob/master/src/gpu/kmeans/kmeans_impl.h#L76)

navdeep-G:
Will add that. Just wanted to fix this slightly in the meantime

Issue #451, opened 2018-02-15
Resolve CRAN CMD check warnings


terrytangyuan:
Resolve warnings from `devtools::check()`

PR #452, opened 2018-02-15
Resolve CRAN check warnings/errors/notes

### LINKED ISSUE #447: Properly skip tests on CRAN
To prepare CRAN submission

terrytangyuan:
This fixes #447 and #451. 

* Resolve CRAN check warnings/errors/notes
* Skip tests on CRAN


navdeep-G:
@terrytangyuan Any idea about the failures? Seems to have happened on this commit: https://github.com/h2oai/h2o4gpu/pull/452/commits/e6f7ddab392112fc5486fcdc3626da72e9e9bccf

terrytangyuan:
I dont't think the failure is related to this PR:
```
Collecting botocore==1.8.43 (from awscli>=1.11.148->-r requirements.txt (line 36))
  Could not find a version that satisfies the requirement botocore==1.8.43 (from awscli>=1.11.148->-r requirements.txt (line 36)) 
No matching distribution found for botocore==1.8.43 (from awscli>=1.11.148->-r requirements.txt (line 36))
make: *** [deps_install] Error 1
Makefile:218: recipe for target 'deps_install' failed
script returned exit code 2
```

navdeep-G:
I've restarted the build

src/interface_r/DESCRIPTION
```diff
@@ -1,13 +1,13 @@
 Package: h2o4gpu
 Type: Package
-Title: h2o4gpu in R
+Title: R Interface to 'H2O4GPU'
```
navdeep-G: Not sure if quotes are needed around H2O4GPU

src/interface_r/DESCRIPTION
```diff
@@ -1,13 +1,13 @@
...
 Version: 0.0.0.9000
 Authors@R: c(
   person("Yuan", "Tang", role = c("aut", "cre"), email = "terry@h2o.ai"),
   person("Navdeep", "Gill", role = c("aut"), email = "navdeep@h2o.ai"),
   person("Erin", "LeDell", role = c("aut"), email = "erin@h2o.ai"),
   person("H2O.ai", role = c("cph", "fnd")))
-Description: h2o4gpu in R
+Description: R Interface to 'H2O4GPU' - A collection of 'GPU' solvers for machine learning algorithms.
```
navdeep-G: Not sure if quotes are needed around H2O4GPU & GPU

terrytangyuan:
That’s a new requirement by CRAN submissions. Any jargons should be surrounded by single quotes.

navdeep-G:
LGTM

ledell:
@terrytangyuan Do you mean that CRAN requires quotes around acronyms?  Is this posted somewhere on CRAN policies that you can link to so I can read more?

src/interface_r/R/auto_generated_wrappers.R
```diff
@@ -378,6 +390,8 @@ h2o4gpu.elastic_net_classifier <- function(
   h2o4gpu_model(model, c("classifier"), "Elastic Net Classifier")
 }
 
+#' @description KMeans Clustering
```
ledell: Is this title generated automatically?  If not, can we change to "K-means"?  (That's the official spelling).

terrytangyuan: I'll change that.

src/interface_r/R/auto_generated_wrappers.R
```diff
@@ -416,6 +430,8 @@ h2o4gpu.kmeans <- function(
   h2o4gpu_model(model, NULL, "KMeans Clustering")
 }
 
+#' @description Principal Components Analysis (PCA)
```
ledell: Correct name is "Principal Component Analysis" (no "s").  Can we change this?  (in Python if necessary)

src/interface_r/R/generics.R
```diff
@@ -21,3 +21,9 @@ fit_transform <- function(object, ...) {
 fit_predict <- function(object, ...) {
   UseMethod("fit_predict")
 }
+
+# Note that we are redefining `transform` generic method here to override `base::transform`'s signature.
+#' @export
+transform <- function(object, ...) {
```
ledell: What does the H2O4GPU `transform()` method do?  I'm curious if we need this or if it's some sklearn artifact that's not necessary in R.

terrytangyuan:
@ledell I was looking for that as well but couldn't find it. It may just be a suggestion that people follow around in many packages. If you run `devtools::release()` you'll see it will do some spelling checks before you submit to CRAN. Other spelling checks in docs are fine but DESCRIPTION file is more strict and it will complain if you don't fix it.

terrytangyuan:
Some examples:

https://cran.r-project.org/web/packages/Xplortext/index.html
https://cran.r-project.org/web/packages/ggplot2/index.html
https://cran.r-project.org/web/packages/tensorflow/index.html
https://cran.r-project.org/web/packages/reticulate/index.html (you need to do this even for "Python")!

src/interface_r/R/generics.R
terrytangyuan: It's usually for unsupervised algorithms but IMO it's the same as `predict()`. @navdeep-G What do you think?

ledell: Ok, if it's just for sklearn and not 100% necessary, then I'd say let's leave it out for now.  We should ship the first version with the bare essentials so we don't end up wanting to remove stuff later. 

I am also thinking we should remove `fit_transform()` and `fit_predict()` and probably `score()`:

- It seems better to follow the standard R convention of separating fit from predict.  
- Do we have any other way to scoring our models right now other than `score()`?  If they use GPU methods, will the sklearn `score()` method work (i think it would just call sklearn to do that underneath)?  We have some options here:
  - We could expect people to use a third-party metrics package (e.g. [Metrics](https://cran.r-project.org/web/packages/Metrics/index.html) is pretty easy to use and comprehensive) to evaluate their models.  
  - We could provide our own built-in metrics functions like we do in H2O-3.
  - We could write a wrapper on top of *Metrics* to give a more integrated feel and return something similar to what we have in H2O-3.
   - Give them `score()` (though that's not very useful since it only provides a single metric, which is usually not the one you want actually -- it's accuracy for classification and R^2 for regression).

We could remove these here or in a follow-up PR.

navdeep-G:
@ledell @terrytangyuan You guys bring up good points. I would like to review this one more time with these points in mind if that’s okay?

terrytangyuan:
Sure take your time. As we get closer to CRAN submission, we really need to think more carefully.

ledell:
@terrytangyuan Let's merge this and then follow up with a discussion about removing the extra functions (e.g. `fit_transform()`, `fit_predict()`, etc) later.

PR #454, opened 2018-02-16
Add iris test for tsvd

navdeep-G:
* Will help debug on realistic datasets for tsvd
* More tests to come...


terrytangyuan:
logging seems to be imported but not used? The print seems very verbose - I also see a lot of print in many other Python files. Are those necessary?

navdeep-G:
Prints are good for tests. We can change to logging but haven’t done so yet.

PR #458, opened 2018-02-23
Make R installation optional based on INSTALL_R env var

terrytangyuan:
This is to address the issue when installing R while trying to make h2o4gpu work for DAI on various native builds. 

* `make fullinstall` should now install R only when `INSTALL_R = 1`.  In other words, `INSTALL_R = 0` is the default so `make fulllinstall` won't install R dependencies now unless you do `make INSTALL_R=1 fullinstall`

* R tests will only run when R is available AND all the necessary R dependencies are available. 

* Allow R version to be conconfigurable prior to installation via `R_VERSION` env var.

### DIFF
--- a/Dockerfile-build
+++ b/Dockerfile-build
@@ -79,8 +79,9 @@ RUN \
     pip install -r requirements_runtime.txt && \
     pip install -r requirements_runtime_demos.txt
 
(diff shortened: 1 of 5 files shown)


navdeep-G:
LGTM

PR #462, opened 2018-02-27
Add float and double variant of tsvd

### LINKED ISSUE #347: Double variant for tSVD
We need to also handle "double" as input type for tSVD (like in KMeans and POGS).

navdeep-G:
* Resolves #347 
* Kept original impl of tsvd for PCA impl. Will change that in a later PR.
* Still cleaning up but wanted to open a PR

### DIFF
--- a/src/gpu/data/matrix.cu
+++ b/src/gpu/data/matrix.cu
@@ -274,7 +274,7 @@ namespace tsvd
 		safe_cusolver(cusolverDnDsyevd_bufferSize(context.cusolver_handle, CUSOLVER_EIG_MODE_VECTOR, CUBLAS_FILL_MODE_UPPER, X.rows(), X.data(), X.columns(), w.data(), &lwork));
 
 		double *d_work;
-		safe_cuda(cudaMalloc(&d_work, sizeof(float) * lwork));
+		safe_cuda(cudaMalloc(&d_work, sizeof(double) * lwork));
 
 		int *dev_info = NULL;
 		safe_cuda(cudaMalloc ((void**)&dev_info, sizeof(int)));
--- a/src/gpu/data/matrix.cu
+++ b/src/gpu/data/matrix.cu
@@ -4,6 +4,49 @@
 
 namespace tsvd
 {
+	template<typename T, typename U>
+	void multiply(Matrix<T>& A, const U a, DeviceContext& context)
+	{
+		thrust::transform(A.dptr(), A.dptr() + A.size(), A.dptr(), [=]__device__ (U val)
+						  {
+							  return val * a;
+						  }
+		);
+	}
+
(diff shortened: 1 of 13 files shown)


src/gpu/data/matrix.cu
```diff
@@ -43,6 +73,15 @@ namespace tsvd
 		);
 	}
 
+	void multiply(Matrix<tsvd_double>& A, const tsvd_double a, DeviceContext& context)
```
mdymczyk: Could these `multiply` methods be generalized into a single templated method, something like:

```
template <typename T>
void multiply(Matrix<T>& A, const T a, DeviceContext& context)
	{
		thrust::transform(A.dptr(), A.dptr() + A.size(), A.dptr(), [=]__device__ (T val)
		                  {
			                  return val * a;
		                  }
		);
	}
```
Or will the compiler complain?

Similar with other methods using `thrust` (add, subtract etc.) - anything not using `cublas`.

mdymczyk:
Added a small comment/question but otherwise looks good 👍

src/gpu/data/matrix.cu
RAMitchell: You can do this it just needs to go in the header file not the .cu file.

RAMitchell:
Looks good at a glance. As @mdymczyk mentioned you can probably remove duplicated functions in a lot of cases. Duplicate code is the root of all evil so worth some time figuring this out :)

Does the increased precision resolve the accuracy issues you were seeing before?

Also note that when you run double precision operations on a GeForce card expect them to be up to 32x slower than single precision operations.

navdeep-G:
Thx for the comments, guys. I will look into removing duplicated code (mainly for thrust calls)

And I have to see about precision. I hope it will help.

navdeep-G:
@mdymczyk @RAMitchell I have added templates to functions using `thrust`. Please see commit with chagne here: https://github.com/h2oai/h2o4gpu/pull/462/commits/00b8a3480bb029f94bba45f7e9970cdd2f54ff94

navdeep-G:
@mdymczyk Not sure, but the logs show this:
```
tests_open/svd/test_tsvd_power.py::test_tsvd_power_k3 Sending interrupt signal to process

./scripts/make-docker-runtests.sh: line 32: 17087 Terminated              $docker exec ${CONTAINER_NAME} bash -c 'export HOME=`pwd`; eval "$(/root/.pyenv/bin/pyenv init -)" ; /root/.pyenv/bin/pyenv global 3.6.1; cd repo ; pip install `find /dot/src/interface_py/'${dist}' -name "*h2o4gpu-*.whl"`; pip freeze ; make '${target}
```
Will rerun as I didn't see any failures.

PR #466, opened 2018-03-04
Add initial AWS instructions

navdeep-G:


mdymczyk:
This isn't really AWS specific, though - just how to install gcc, make and Python on Ubuntu and as I mentioned in the issue repeats a fair bit of information already provided in the README. This might be a bit confusing for the user also, as certain AMIs don't use Ubuntu - we got a comment in one of the issues yesterday where someone was using the Amazon distro and this would not work.

What I think would be more robust and easy to maintain for us would be to add links to external resources e.g. mention in the requirements (README) we need Python and link it to http://docs.python-guide.org/en/latest/starting/installation/ which contains installation instructions for all systems. Similar thing can be done for gcc and make.

navdeep-G:
Yes. Will close

Issue #477, opened 2018-03-12
Fix R CMD check warnings


terrytangyuan:
All warnings need to be fixed before submitting to CRAN. Here's the current output of R CMD check:

```
==> devtools::check(document = FALSE)

Setting env vars --------------------------------------------------------------
CFLAGS  : -Wall -pedantic
CXXFLAGS: -Wall -pedantic
Building h2o4gpu --------------------------------------------------------------
'/usr/lib/R/bin/R' --no-site-file --no-environ --no-save --no-restore --quiet  \
  CMD build '/home/terry/h2o4gpu/src/interface_r' --no-resave-data  \
  --no-manual 

* checking for file ‘/home/terry/h2o4gpu/src/interface_r/DESCRIPTION’ ... OK
* preparing ‘h2o4gpu’:
* checking DESCRIPTION meta-information ... OK
Warning: /tmp/RtmpxKwpRA/Rbuild2d023b84719d/h2o4gpu/man/h2o4gpu.pca.Rd:20: unknown macro '\item'
Warning: /tmp/RtmpxKwpRA/Rbuild2d023b84719d/h2o4gpu/man/h2o4gpu.pca.Rd:22: unknown macro '\item'
Warning: /tmp/RtmpxKwpRA/Rbuild2d023b84719d/h2o4gpu/man/h2o4gpu.pca.Rd:24: unknown macro '\item'
Warning: /tmp/RtmpxKwpRA/Rbuild2d023b84719d/h2o4gpu/man/h2o4gpu.pca.Rd:26: unknown macro '\item'
Warning: /tmp/RtmpxKwpRA/Rbuild2d023b84719d/h2o4gpu/man/h2o4gpu.pca.Rd:28: unknown macro '\item'
Warning: /tmp/RtmpxKwpRA/Rbuild2d023b84719d/h2o4gpu/man/h2o4gpu.pca.Rd:30: unexpected section header '\description'
Warning: /tmp/RtmpxKwpRA/Rbuild2d023b84719d/h2o4gpu/man/h2o4gpu.pca.Rd:33: unexpected END_OF_INPUT '
'
* checking for LF line-endings in source and make files
* checking for empty or unneeded directories
* building ‘h2o4gpu_0.0.0.9000.tar.gz’

Setting env vars --------------------------------------------------------------
_R_CHECK_CRAN_INCOMING_USE_ASPELL_: TRUE
_R_CHECK_CRAN_INCOMING_           : FALSE
[...]

PR #478, opened 2018-03-12
Add documentation for generic methods for R API

terrytangyuan:
This fixed some of the CMD check warnings in #477, e.g. 

```
* checking for missing documentation entries ... WARNING
Undocumented code objects:
  ‘fit’ ‘fit_predict’ ‘fit_transform’ ‘h2o4gpu_model’ ‘score’
  ‘transform’
```

Remove some methods since users can always chain methods ourselves.
@ledell I removed `score()` since we decided to use third-party R libraries for calculating performance metrics.


src/interface_r/scripts/gen_wrappers.R
```diff
@@ -13,8 +13,8 @@ pkg_location <- "~/h2o4gpu/src/interface_r"
 
 
 # Note: Code below should not be changed unless necessary
-
-source(file.path(pkg_location, "scripts/gen_wrapper_utils.R"))
+setwd(pkg_location)
```
terrytangyuan: This is needed because the file names are all based on it. I think it's safe to change working directory to this (also makes things easier for development).

navdeep-G:
LGTM

PR #480, opened 2018-03-13
daal library installation within miniConda

stjoern:
daal installation with miniConda on Centos platform


stjoern:
for manual daal installation, please make cpu-fullinstall, there is a script behind for downloading pyDaal wheel. Currently this build should install dall on Centos-x86-64 architecture, I still don't have access to AIX machines to build it there.

stjoern:
- fixed error in RR
- SVD and unit tests added

PR #481, opened 2018-03-14
Fixes #457: Add R package install instructions in the main README.md

### LINKED ISSUE #457: Update README.md to include R installation
Once we're ready to release the R package for the first time, let's add some instructions to the main `README.md` file for this repo, to tell users how to install in R.  We can keep this short and link to to the R-specfic [README.md](https://github.com/h2oai/h2o4gpu/blob/master/src/interface_r/README.md) for more detailed instructions.

terrytangyuan:
Per request in #457:

* Updated example in README
* Added instructions in main README and reorganized a bit

We'll add CRAN installation instruction later too once on CRAN.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,15 +2,8 @@
 
 [![Join the chat at https://gitter.im/h2oai/h2o4gpu](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/h2oai/h2o4gpu)
 
-**H2O4GPU** is a collection of GPU solvers by
-[H2Oai](https://www.h2o.ai/).  It builds upon the easy-to-use
-[Scikit-learn](http://scikit-learn.org) API and its well-tested
-CPU-based algorithms.  It can be used as a drop-in replacement for
-scikit-learn (i.e. `import h2o4gpu as sklearn`) with support for GPUs
-on selected (and ever-growing) algorithms.  H2O4GPU inherits all the
-existing scikit-learn algorithms and falls-back to CPU aglorithms when
-the GPU algorithm does not support an important existing Scikit-learn
-class option.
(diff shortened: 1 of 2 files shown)


navdeep-G:
LGTM

README.md
```diff
@@ -110,14 +119,32 @@ array([[ 1.,  1.  ],
        [ 1.,  4.  ]])
 ```
 
-For more examples check our [Jupyter notebook demos](https://github.com/h2oai/h2o4gpu/tree/master/examples/py/demos).
+To test your installation of the R package, try the following example that builds a simple [XGBoost](https://github.com/dmlc/xgboost) random forest classifier:
+
+``` r
+require(h2o4gpu)
```
ledell: should be `library(h2o4gpu)`.  They do the same thing, but `require()` is designed only for use within other functions.

ledell:
On this PR, we could also add some more info in the main H2O4GPU description about the R package as well.  Here's what we have now:

**H2O4GPU** is a collection of GPU solvers by
[H2Oai](https://www.h2o.ai/).  It builds upon the easy-to-use
[Scikit-learn](http://scikit-learn.org) API and its well-tested
CPU-based algorithms.  It can be used as a drop-in replacement for
scikit-learn (i.e. `import h2o4gpu as sklearn`) with support for GPUs
on selected (and ever-growing) algorithms.  H2O4GPU inherits all the
existing scikit-learn algorithms and falls-back to CPU aglorithms when
the GPU algorithm does not support an important existing Scikit-learn
class option.

Here are proposed edits (let me know if this looks good and I can commit, or suggest changes) to this PR:

**H2O4GPU** is a collection of GPU solvers by [H2Oai](https://www.h2o.ai/) with APIs in Python and R.  The Python API builds upon the easy-to-use [Scikit-learn](http://scikit-learn.org) API and its well-tested CPU-based algorithms.  It can be used as a drop-in replacement for
scikit-learn (i.e. `import h2o4gpu as sklearn`) with support for GPUs
on selected (and ever-growing) algorithms.  H2O4GPU inherits all the
existing scikit-learn algorithms and falls-back to CPU algorithms when
the GPU algorithm does not support an important existing scikit-learn
class option.  The R package is a wrapper around the H2O4GPU Python package, and the interface follows standard R conventions for modeling.

terrytangyuan:
Looks good to me. I can change it. Though should we duplicate the description in both READMEs?

ledell:
@terrytangyuan Do you mean duplicate this paragraph to the [`src/interface_r/README.md`](https://github.com/h2oai/h2o4gpu/tree/master/src/interface_r)?  Sure, we could do that (right now the R README.md only has short description).  I also noticed that we use `score()` in the R README, so maybe you could delete that for now as well.

terrytangyuan:
Yes. @ledell I just updated the PR to address your comments. Thanks!

PR #483, opened 2018-03-15
Utilize Python for explained variance calc in tsvd and pca

### LINKED ISSUE #450: Possible bug in explained variance/ explained variance ratio
The differences I am seeing with explained variance and explained variance ratio between CUDA and Python. Done on a 5000 by 50 matrix with k = 2:
Calculation in Python
```
h2o4gpu tsvd Explained Variance
[0.08407421 0.09914753]
Sklearn Explained Variance
[0.08407422 0.09914825]

```
Calculation in CUDA:
```
h2o4gpu tsvd Explained Variance
[0.08407421 0.08716473]
Sklearn Explained Variance
[0.08407422 0.09914825]
```

navdeep-G:
* Resolves #450

### DIFF
--- a/src/interface_py/h2o4gpu/solvers/truncated_svd.py
+++ b/src/interface_py/h2o4gpu/solvers/truncated_svd.py
@@ -178,15 +178,13 @@ class TruncatedSVDH2O(object):
         full_var = \
             np.var(X, axis=0).sum()
         if self.verbose:
-            print("Time taken for full variance : "
-            + str(time.time() - start_var))
+            print("Time taken for full variance : " + str(time.time() - start_var))
         if self.verbose:
             start_evr = time.time()
         self.explained_variance_ratio = \
             self.explained_variance / full_var
         if self.verbose:
-            print("Time taken for explained variance ratio : "
-            + str(time.time() - start_evr))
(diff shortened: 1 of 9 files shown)


src/interface_py/h2o4gpu/solvers/truncated_svd.py
```diff
@@ -157,41 +155,36 @@ def fit_transform(self, X, y=None):
...
+        full_var = \
+            np.var(X, axis=0).sum()
+        if self.verbose:
+            print("Time taken for full variance : " + str(time.time() - start_var))
+        if self.verbose:
+            start_evr = time.time()
+        self.explained_variance_ratio = \
+            self.explained_variance / full_var
+        if self.verbose:
+            print("Time taken for explained variance ratio : " + str(time.time() - start_evr))
```
terrytangyuan: Seems like a great place to use `logging` instead of `print`.

mdymczyk:
So is #450 due to CUDA's numerical (float) instability and works fine for `double` (in tSVD) or is it just a bug on our end we cannot find? 

Not a huge performance hit I guess, just want to find out the cause.

navdeep-G:
@mdymczyk The latter. Can't find the source and want to get this potential fix in as it doesn't burden perf.

PR #487, opened 2018-03-20
fix - performance issues

stjoern:
Jenkins build takes too much time, solving performance issues.


terrytangyuan:
I think we should remove commented code though.

Issue #433, opened 2018-02-01
Add missing `predict_proba` and `predict_log_proba` generic methods when applicable


terrytangyuan:
Add them as generic methods for now. We can also consider a single `predict()` function with argument `type = c("probability", "raw", "log_probability")` that is more old R fashioned. @ledell @navdeep-G Any preference here?

ledell:
I think we can skip the `predict_proba()` and `predict_log_proba()` functions in the R API (at least directly exposed to the user).  

We should definitely support `predict()` and so we'd just have to decide on what we'd like to output from that.  

1. We could follow the H2O convention of returning:
- k+1 columns for k-class classification (first column is predicted class, additional columns are predicted values for each class).
- single column for regression (just the raw predicted value).

2. Or use the `type` argument to produce different results. 
- in R, usually `type = "response"` gives raw predicted values (or classes) and `"prob"` gives a matrix of "probabiliities" though unlike sklearn's `predict_proba()` which uses Platt scaling, these are not really probabilities, so the name is not great.  So one option is to allow types `"response"` and `"prob"`.
- One concern I have about using `predict_proba()` underneath to produce probabilities is that it only works for sklearn models which is probably not what people will be using 95% of the time, and also requires [much more computation](https://stackoverflow.com/a/15112563/5451344).  So my opinion would be not enable a `type` option which attempts to call either `predict_proba()` `predict_log_proba()` underneath.  

Here's a few examples from R for demonstration:

Classification:
```
> library(randomForest)
> rf <- randomForest(x = iris[,1:4], y = iris[,5])
> head(predict(rf, iris, type = "response"))
     1      2      3      4      5      6 
setosa setosa setosa setosa setosa setosa 
Levels: setosa versicolor virginica
> head(predict(rf, iris, type = "prob"))
  setosa versicolor virginica
1  1.000      0.000         0
2  0.998      0.002         0
3  1.000      0.000         0
4  1.000      0.000         0
5  1.000      0.000         0
6  1.000      0.000         0
```

Regression:
```
> rf <- randomForest(x = iris[,1:3], y = iris[,4])
> head(predict(rf, iris, type = "response"))
        1         2         3         4         5         6 
0.2519586 0.2026082 0.2093314 0.1946441 0.2344817 0.3564690 
> head(predict(rf, iris, type = "prob"))
Error in predict.randomForest(rf, iris, type = "prob") : 
  'prob' or 'vote' not meaningful for regression
```
[...]

PR #489, opened 2018-03-21
Modify predict method in R to account for probabilities and update predict_proba in Py API

navdeep-G:
* Resolves https://github.com/h2oai/h2o4gpu/issues/433 by adding a `type` argument to `predict()` with values of "raw" (default; vector of predictions) or "prob" (matrix of uncalibrated class probabilities).  The values of "raw" and "prob" come from the caret R package.
* Sample output in R:
```{r}
predictions <- model %>% predict(x, type = "prob")
head(predictions)
             0         1
[1,] 0.3294092 0.6705908
[2,] 0.3176789 0.6823211
[3,] 0.3176789 0.6823211
[4,] 0.4417148 0.5582852
[5,] 0.4781650 0.5218350
[6,] 0.6315404 0.3684596
```

* New `predict` API:
```{r}
#' Make Predictions using Trained H2O4GPU Estimator
#' 
#' This function makes predictions from new data using a trained H2O4GPU model and returns class predictions
#' for classification and predicted values for regression.
#' 
#' @param object The h2o4gpu model object
#' @param x The new data where each column represents a different predictor variable to 
#' be used in generating predictions.
#' @param type One of "raw" or "prob", indicating the type of output: predicted values or probabilities
#' @param ... Additional arguments (unused for now).
#' @export
predict.h2o4gpu_model <- function(object, x, type="raw", ...) {
  if (type == "raw") {
    preds <- object$model$predict(X = resolve_model_input(x), ...)
  } else if (type == "prob") {
    preds <- object$model$predict_proba(X = resolve_model_input(x), ...)
    if (!is.null(object$classes_)){
      colnames(preds) <- object$classes_ #Taken from tree based models
    }
  } else {
    stop(paste0("Unrecognized 'type' parameter value. Expected either 'raw' or 'prob but got ", type))
  }
  return(preds)
}
```

### DIFF
--- a/src/interface_r/R/model.R
+++ b/src/interface_r/R/model.R
@@ -92,12 +92,16 @@ fit.h2o4gpu_model <- function(object, x, y = NULL, ...) {
 #' @export
 predict.h2o4gpu_model <- function(object, x, type="raw", ...) {
   if (type == "raw") {
-    object$model$predict(X = resolve_model_input(x), ...)
+    preds <- object$model$predict(X = resolve_model_input(x), ...)
   } else if (type == "prob") {
-    object$model$predict_proba(X = resolve_model_input(x), ...)
+    preds <- object$model$predict_proba(X = resolve_model_input(x), ...)
+    if (!is.null(object$classes_)){
+      colnames(preds) <- object$classes_ #Taken from tree based models
+    }
   } else {
     stop(paste0("Unrecognized 'type' parameter value. Expected either 'raw' or 'prob but got ", type))
   }
+  return(preds)
 }
 
 #' Transform a Dataset using Trained H2O4GPU Estimator
--- a/src/interface_r/R/model.R
+++ b/src/interface_r/R/model.R
@@ -81,15 +81,23 @@ fit.h2o4gpu_model <- function(object, x, y = NULL, ...) {
 
 #' Make Predictions using Trained H2O4GPU Estimator
 #' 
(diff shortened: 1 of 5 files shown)


src/interface_r/R/model.R
```diff
@@ -81,15 +81,27 @@ fit.h2o4gpu_model <- function(object, x, y = NULL, ...) {
...
+predict.h2o4gpu_model <- function(object, x, type="raw", ...) {
+  if (type == "raw") {
+    preds <- object$model$predict(X = resolve_model_input(x), ...)
+  } else if (type == "prob") {
+    preds <- object$model$predict_proba(X = resolve_model_input(x), ...)
+    if (!is.null(object$classes_)){
+      colnames(preds) <- object$classes_ #Taken from tree based models
+    }
+  } else {
+    stop(paste0("Unrecognized 'type' parameter value. Expected either 'raw' or 'prob but got ", type))
```
terrytangyuan: Instead of doing this, you can do type=c(“raw”, “prob”) in the arg and then use match.arg() to validate it. This is fine too though.

ledell: Yeah, good call, @terrytangyuan.  Let's fix this in a follow-up PR so that we can merge now and I can use this in the vignette right away instead of waiting for all the tests to re-run.

terrytangyuan: Oops sorry. I thought I already approved it.

PR #497, opened 2018-03-22
Fix py modules loading in R API

terrytangyuan:
Fix the issue “attempt to apply non-function”. This is because those Python modules are not loaded properly.


navdeep-G:
LGTM

PR #500, opened 2018-03-22
Fix version number in Python

mdymczyk:
`__version__` hardcoded to 0.0.4, this should ideally be taken from BUILD_INTO.txt or __about__.py (#501) but before the release let's just swap it to 0.2.0.


navdeep-G:
LGTM

PR #502, opened 2018-03-22
[WIP] CMake rewrite

mdymczyk:


mdymczyk:
@pseudotensor @navdeep-G the build seems to be passing, think I'm done. Can I get a final check before merge?

@pseudotensor I went through the things we found during our review and fixed most:

* DEBUG AND USENVTX are back (with description in DEVEL.md how to use)
* Added an example how to add new code, it's in EXAMPLE_SOLVER.md
* removed a lot of redundant declarations in *.i files (not all) by using header files - unfortunately looks like SWIG requires to have those declarations in 2 places so that's the only way to DRY it
* changed the `int *` to `size_t *` where appropriate

2 more things I need to double check:

1) SWIG generated wrapper complains about `ch2o4gpu_gpu.py:15: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses`
2) need to make sure SWIG isn't creating any new pointers/objects and we don't have any leaks - any ideas how to make an automatic test for this?

navdeep-G:
LGTM. Will test out more once in master. :)

PR #507, opened 2018-03-23
F0.5, F2 and Accuracy for daicx

Far0n:
C++ implementation of optimizing F0.5, F2 and Accuracy with python bindings.

### DIFF
--- a/src/cpu/metrics/metrics.cpp
+++ b/src/cpu/metrics/metrics.cpp
@@ -29,9 +29,24 @@ namespace h2o4gpu {
     return (std::abs(y) < 1E-15) ?  0.0 : x / y;
   }
 
+  double f05(double tp, double tn, double fp, double fn) {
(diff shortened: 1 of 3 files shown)


mdymczyk:
@Far0n looks good! Could we get some Python unit tests?

pseudotensor:
Yes, unit tests would be good.  Let's do that as a next pass since this is for fixing a critical DAI issue.

Far0n:
@mdymczyk you mean directly in h2o4gpu? currently the tests are in DAI, because the code is only used there.

mdymczyk:
@Far0n yes if we could have some tests here it would be great, having tests in DAI is better than nothing but might be a bit too late to find potential issues ;-)

Far0n:
@mdymczyk that's true. :) I will deliver in addition.

PR #509, opened 2018-03-23
Fixed #508

### LINKED ISSUE #508: Update GLM Demo to use predict_proba
In the H2O4GPU_GLM demo, the predictions are calculated by calling predict on the GLM model.  This returns the class predictions, not the probability predictions, and causes the calculated logloss to be Inf.  

Change the demo to return the predicted probability instead of predicted class.

meganjkurka:
- returns the predicted probabilities instead of predicted class in the GLM demo


navdeep-G:
LGTM

PR #538, opened 2018-04-06
Fix GPU usage so doesn't lock on py3nvml import

pseudotensor:


src/gpu/utils.cu
```diff
@@ -29,11 +29,22 @@ extern "C" {
...
   #ifdef __cplusplus
 }
 #endif
 
+#ifdef __cplusplus
+extern "C" {
+  #endif
+
+  int cudaresetdevice_bare(void) {
+    cudaDeviceReset();
```
mdymczyk: Should we do a CUDACHECK() here?

src/interface_py/h2o4gpu/util/gpu.py
```diff
@@ -54,15 +56,20 @@ def get_gpu_info(return_usage=False):
...
-            future = executor.submit(get_gpu_info_subprocess, return_usage)
-            res = future.result()
-        return res
-    except concurrent.futures.process.BrokenProcessPool:
-        if return_usage:
-            return (total_gpus, total_mem, gpu_type, usage)
-        return (total_gpus, total_mem, gpu_type)
+    for trial in range(0, trials):
+        try:
+            with ProcessPoolExecutor(max_workers=1) as executor:
```
mdymczyk: I was wondering for some time - why are we doing this with an executor? Can't we just call `get_gpu_info`? Since we are doing `future.result` anyway which I assume is synchronous? Wouldn't that fix the issue?

Ah and a minor one - pylint is complaining.

src/interface_py/h2o4gpu/util/gpu.py
```diff
@@ -149,6 +156,18 @@ def cudaresetdevice(gpu_id, n_gpus):
         from ctypes import c_int
         lib.cudaresetdevice(c_int(gpu_id), c_int(n_gpus))
 
+def cudaresetdevice_bare(n_gpus):
```
mdymczyk: Minor nitpick - could we rename it from `_bare` to something like `_current` (or something describing the fact it will reset the current process' device)?

src/gpu/utils.cu
mdymczyk: Should we do a `CUDACHECK` here? Also probably we can return `void` here instead?

src/interface_py/h2o4gpu/util/gpu.py
```diff
@@ -54,15 +56,20 @@ def get_gpu_info(return_usage=False):
...
-    except concurrent.futures.process.BrokenProcessPool:
-        if return_usage:
-            return (total_gpus, total_mem, gpu_type, usage)
-        return (total_gpus, total_mem, gpu_type)
+    for trial in range(0, trials):
+        try:
+            with ProcessPoolExecutor(max_workers=1) as executor:
+                future = executor.submit(get_gpu_info_subprocess, return_usage)
+                # don't wait more than 30s, import on py3nvml can hang if 2 subprocesses GIL lock import at same time
+                res = future.result(timeout=timeout)
```
mdymczyk: I've been wondering why are we doing this with a subprocess? Can't we just call `get_gpu_info_subprocess` since we call `result` (which I assume is synchronous anyway) but we expose ourselves to potential deadlocks/race conditions? Wouldn't that fix the problem?

Or at least we can add a comment why are we doing it this way.

src/gpu/utils.cu
pseudotensor: No, can't, because that would fail as no cuda context or it would regenerate the cuda context.  We want to destroy it and be done.

src/interface_py/h2o4gpu/util/gpu.py
pseudotensor: I mention why.  py3nvml can create a cuda context and one can't share those across into a child process.  So we have to isolate any cuda/nvml calls out of the parent.

src/interface_py/h2o4gpu/util/gpu.py
pseudotensor: It's _bare() in sense that no checks or other cuda calls.  Needs to be bare bones.

pseudotensor:
@mdymczyk What did you mean by re-use existing methods?    I don't follow.

src/gpu/utils.cu
mdymczyk: @pseudotensor ah I see, so in `cudaresetdevice` we can do it because we are doing an explicit `cudaSetDevice`? Could we add a comment here so I don't add any checks in the future by accident? (during last refactoring I noticed we have a few cuda check macros/methods - would be good to unify that).

src/interface_py/h2o4gpu/util/gpu.py
mdymczyk: Uh yeah my bad for not reading the docstring, sorry.

mdymczyk:
@pseudotensor thought we can reuse `cudaresetdevice` but guess not since we can't do check. Looks good in that case 👍

PR #542, opened 2018-04-10
Pip >10.0 compatibility fix.

mdymczyk:
Pip >10 doesn't support pip.req anymore


navdeep-G:
LGTM. Lets see how the build goes.

PR #546, opened 2018-04-11
Metrics SWIG inclusion.

mdymczyk:
Include metrics.i in our main interface.

Return value from conf matrix method.


Far0n:
@mdymczyk do we need to put the **numpy.i.** (https://github.com/h2oai/h2o4gpu/blob/master/src/interface_py/h2o4gpu/util/numpy.i elsewhere 2)?

mdymczyk:
@Far0n no we don't need that one at all, I just forgot to put `metrics.i` into our (now) main interface file during cmake/swig remake - my miss sorry.

Far0n:
@mdymczyk my code uses **numpy.i**

Far0n:
@mdymczyk Why has **metrics.i** been changed? I think its broken now. Someone removed the **numpi.i** inclusion.

mdymczyk:
@Far0n we're using now SWIG for everything, its all in here now https://github.com/h2oai/h2o4gpu/tree/master/src/swig (together with numpy.i).

Far0n:
@mdymczyk I created *metrics.i* initially and it uses numpy makros:

```
%apply (double* IN_ARRAY1, int DIM1) {(double *y, int n),
                                      (double *yhat, int m),
                                      (double *w, int l)};
%apply (double* INPLACE_ARRAY2, int DIM1, int DIM2) {(double *cm, int k, int j)};
```
but **numpi.i** is no longer included. Don't think DAICX-code compiles any longer or how does that work?

mdymczyk:
@Far0n we have now 1 main interface file per CPU and GPU lib (e.g. https://github.com/h2oai/h2o4gpu/blob/bugfix/include-metrics-swig/src/swig/ch2o4gpu_cpu.i), inside which we include:

- `numpy.i`
- all of our own interfaces, including your `metrics.i`

We don't need to include `numpy.i` in `metrics.i` as it's already included in the main interface. 

When refactoring I forgot to include `metrics.i`, hence the DAI error. Should be fine now.

Far0n:
@mdymczyk that's cool. sorry for confusion, missed that one. Nice to see more SWIG used. :)

PR #555, opened 2018-04-13
Add CUDA version of whiten parameter for PCA

navdeep-G:
* Resolves https://github.com/h2oai/h2o4gpu/issues/552

### DIFF
--- a/src/gpu/tsvd/tsvd.cu
+++ b/src/gpu/tsvd/tsvd.cu
@@ -160,7 +160,6 @@ namespace tsvd
 				d_u[idx] = 0.0;
 			}
 		} );
-
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


src/gpu/tsvd/tsvd.cu
```diff
@@ -437,11 +452,11 @@ namespace tsvd
...
-template void tsvd::truncated_svd_matrix<double>(Matrix<double> &X, double* _Q, double* _w, double* _U, double* _explained_variance, double* _explained_variance_ratio, params _param);
-template void tsvd::cusolver_tsvd<double>(Matrix<double> &X, double* _Q, double* _w, double* _U, double* _explained_variance, double* _explained_variance_ratio, params _param);
-template void tsvd::power_tsvd<double>(Matrix<double> &X, double* _Q, double* _w, double* _U, double* _explained_variance, double* _explained_variance_ratio, params _param);
-template void tsvd::cusolver_tsvd<float>(Matrix<float> &X, float* _Q, float* _w, float* _U, float* _explained_variance, float* _explained_variance_ratio, params _param);
-template void tsvd::power_tsvd<float>(Matrix<float> &X, float* _Q, float* _w, float* _U, float* _explained_variance, float* _explained_variance_ratio, params _param);
+template void tsvd::truncated_svd_matrix<float>(Matrix<float> &X, float* _Q, float* _w, float* _U, float* _explained_variance, float* _explained_variance_ratio, params _param, bool _whiten);
```
mdymczyk: Lets put this `_whiten` in `tsvd::params` like we do in `pca::params` - will require less method signature changes and a more condense parameter list.

src/gpu/tsvd/tsvd.cu
```diff
@@ -171,14 +170,30 @@ namespace tsvd
...
+			auto counting = thrust::make_counting_iterator <int>(0);
+			thrust::for_each(counting, counting+QtTrunc.size(), [=]__device__(int idx){
+				int column = idx/column_size;
+				float sigma = d_sigma[column];
+				float q = d_q[idx];
+				d_q[idx] = (q * x_sqrt_row)/std::sqrt(sigma);
+			} );
+			QtTrunc.copy_to_host(_Q); //Send to host
+		} else {
+			QtTrunc.copy_to_host(_Q); //Send to host
```
mdymczyk: Don't need this `else`, just do `QtTrunc.copy_to_host(_Q);` always (and remove it from the `if(_whiten)`).

src/gpu/tsvd/tsvd.cu
```diff
@@ -171,14 +170,30 @@ namespace tsvd
...
+		if (_whiten) {
+			auto d_q = QtTrunc.data();
+			auto x_sqrt_row = std::sqrt(X.rows());
+			auto d_sigma = w.data();
+			auto column_size = QtTrunc.rows();
+			auto counting = thrust::make_counting_iterator <int>(0);
+			thrust::for_each(counting, counting+QtTrunc.size(), [=]__device__(int idx){
+				int column = idx/column_size;
+				float sigma = d_sigma[column];
+				float q = d_q[idx];
```
mdymczyk: Shouldn't `sigma` and `q` be of type `T`?

mdymczyk: Noticed we do similar things in other methods e.g. `calculate_u` - think quite a few params should be changed from `float` (also double? not sure about those) to `T` or we'll be loosing precision?

navdeep-G:
@mdymczyk I have updated the impl based on your comments.

src/interface_py/h2o4gpu/solvers/truncated_svd.py
```diff
@@ -124,6 +124,7 @@ def fit_transform(self, X, y=None):
         param.random_state = self.random_state
         param.verbose = self.verbose
         param.gpu_id = self.gpu_id
+        param.whiten = False #Whitening is not exposed for tsvd yet
```
navdeep-G: I might/might not expose this parameter. Just want to leave here as a reminder for myself.

Issue #514, opened 2018-03-24
Add util to prep data for h2o4gpu python package


navdeep-G:

ledell:
@navdeep-G What kind of data prep?  I assume you mean things like label encoding, one-hot encoding & imputation?  Can we just expose the Scikit-learn methods for this, or do we need to write new methods from scratch?

navdeep-G:
I was thinking of utilizing py datatable somehow.

ledell:
@navdeep-G It would be ideal to use py datatable on the backend, while keeping the scikit-learn API.  Here's a list of all the scikit-learn preprocessing methods that we could use (and fall back to scikit-learn when GPU not supported?): http://scikit-learn.org/stable/modules/classes.html#module-sklearn.preprocessing

navdeep-G:
Yes, but both operations are CPU. So no need to fallback.

ledell:
@navdeep-G I thought py datatable was GPU capable?

navdeep-G:
@ledell py datatable is pure CPU as of today.

navdeep-G:
Researching more and I think sklearn's capabilities achieve what is needed and most users of Python machine learning libraries will know how to use these methods.

PR #566, opened 2018-04-19
Add util to prep data for h2o4gpu R package

### LINKED ISSUE #513: Add util to prep data for h2o4gpu R package
* Data needs to be numeric for all algos

navdeep-G:
* Resolves #513
* This will act as a guide for the Python impl (#514)


src/interface_r/R/data_prep.R
```diff
@@ -0,0 +1,102 @@
...
+#'@param data_table `data.table` object containing data that needs to be preprocessed for h2o4gpu
+#'@param data_path Path to data that needs to be preprocessed for h2o4gpu. Only needed if `data_table` parameter is not provided/available
+#'@param response Response column as a string or index
+#'@param save_as_csv Whether to save processed data as a csv
+#'@param save_csv_path Path to save processed data as a csv
+#'@export
+prep_data <- function(data_table=NULL, data_path=NULL, response=NULL, save_as_csv=FALSE, save_csv_path=NULL){
+  
+  if (save_as_csv && is.null(save_csv_path)) {
+    stop("`save_as_csv` is set to TRUE but `save_csv_path` is NULL. Please specify a path to save processed data")
```
terrytangyuan: you can combine these two parameters into one `save_csv_path`. If it's not NULL, then save it.

src/interface_r/R/data_prep.R
```diff
@@ -0,0 +1,102 @@
...
+      DT <- fread(data_path)
+    } else {
+      stop("Parameters `data_table` & `data_path` are NULL. Need to specify at least one.")
+    }
+  } else {
+    DT <- data_table
+  }
+  
+  if (!is.null(response)) {
+    response <- response
```
terrytangyuan: typo?

src/interface_r/R/data_prep.R
```diff
@@ -0,0 +1,102 @@
...
+    if (is.character(response)) {
+      print(paste0("Response is -> ",response))
+    } else {
+      print(paste0("Response is -> ",colnames(DT)[response]))
+    }
+  } else {
+    stop("Response is not specified")
+  }
```
terrytangyuan: you can simply remove the default value of this argument.

src/interface_r/R/data_prep.R
```diff
@@ -0,0 +1,102 @@
...
+  print(paste0("Number of columns: ", ncol(DT)))
+  
+  print(paste0("Number of rows: ", nrow(DT)))
+
+  ## Label-encoding of categoricals (those cols with fewer than 1k levels, but not constant)
+  print("Label encoding dataset...")
+  feature.names <- setdiff(names(DT), response)
+  for (ff in feature.names) {
+    tt <- uniqueN(DT[[ff]])
+    if (tt < 1000 && tt > 1) {
```
terrytangyuan: why 1k? This should be documented in docstring as well

src/interface_r/R/data_prep.R
```diff
@@ -0,0 +1,102 @@
...
+      DT[, (ff):=factor(DT[[ff]])]  
+      print(paste0(ff," has ",tt," levels"))
```
terrytangyuan: I am concerned about the verbosity of this function. Should we have an `verbose` argument?

src/interface_r/R/data_prep.R
```diff
@@ -0,0 +1,102 @@
...
+  
+  print(paste0("Number of columns after label encoding: ", ncol(DT)))
+  
+  numCols <- names(DT)[which(sapply(DT, is.numeric))]
+  catCols <- names(DT)[which(sapply(DT, is.factor))]
+  print(paste0("Number of numeric columns: ",     length(numCols)))
+  print(paste0("Number of categorical columns: ", length(catCols)))
+  
+  ## impute missing values, drop near-const cols and standardize the data
+  print("Imputing missing values using mean...")
```
terrytangyuan: why using mean? I'd imagine users want to do custom imputation so maybe we should expose this somehow

src/interface_r/R/data_prep.R
navdeep-G: Yep, can do that.

src/interface_r/R/data_prep.R
navdeep-G: Not a typo.

src/interface_r/R/data_prep.R
terrytangyuan: for example, they want to pass their own centers and scales or totally different imputation algorithm

src/interface_r/R/data_prep.R
navdeep-G: I prefer to keep as NULL if thats ok. Unless you  mean something else?

src/interface_r/R/data_prep.R
navdeep-G: No reason for 1K. This is just a number we have used internally before.

src/interface_r/R/data_prep.R
```diff
@@ -0,0 +1,102 @@
...
+  print(paste0("Number of columns after mean imputation: ", ncol(DT)))
+  
+  ## one-hot encode the categoricals
+  print("One hot encoding data table categoricals only...")
+  DT2 <- as.data.table(model.matrix(DT[[response]]~., data = DT[,c(catCols), with=FALSE], sparse=FALSE))[,-1]
+  print(paste0("Number of columns that have been one hot encoded: ", ncol(DT2)))
+  
+  ## add back the numeric columns and assign back to DT
+  print("Add back numeric columns and assign to data table")
+  DT <- DT2[,(numCols):=DT[,numCols,with=FALSE]]
```
terrytangyuan: so this requires two copies of the datasets in memory?

src/interface_r/R/data_prep.R
navdeep-G: Yes, we can expose it. This is just a first iteration.

src/interface_r/R/data_prep.R
navdeep-G: I think I can overwrite.

src/interface_r/R/data_prep.R
navdeep-G: We can add verbose.

src/interface_r/R/data_prep.R
navdeep-G: `DT2` contains the one hot encoded variables.

src/interface_r/R/data_prep.R
```diff
@@ -0,0 +1,102 @@
...
+      print(paste0("Dropping constant column: ", ff))
+      DT[, (ff):=NULL]
+    }
+  }
+  
+  print(paste0("Number of columns after label encoding: ", ncol(DT)))
+  
+  numCols <- names(DT)[which(sapply(DT, is.numeric))]
+  catCols <- names(DT)[which(sapply(DT, is.factor))]
+  print(paste0("Number of numeric columns: ",     length(numCols)))
```
terrytangyuan: redundant space?

src/interface_r/R/data_prep.R
```diff
@@ -0,0 +1,102 @@
+#'Prep data for h2o4gpu algorithms
+#'
+#'@param data_table `data.table` object containing data that needs to be preprocessed for h2o4gpu
+#'@param data_path Path to data that needs to be preprocessed for h2o4gpu. Only needed if `data_table` parameter is not provided/available
+#'@param response Response column as a string or index
+#'@param save_as_csv Whether to save processed data as a csv
+#'@param save_csv_path Path to save processed data as a csv
+#'@export
+prep_data <- function(data_table=NULL, data_path=NULL, response=NULL, save_as_csv=FALSE, save_csv_path=NULL){
```
terrytangyuan: spaces between equal signs (here and in some other places)?

src/interface_r/R/data_prep.R
navdeep-G: yes

src/interface_r/R/data_prep.R
navdeep-G: I can space them out.

src/interface_r/R/data_prep.R
```diff
@@ -0,0 +1,102 @@
...
+  
+  if (save_as_csv && is.null(save_csv_path)) {
+    stop("`save_as_csv` is set to TRUE but `save_csv_path` is NULL. Please specify a path to save processed data")
+  }
+  
+  if (is.null(data_table)) {
+    if (!is.null(data_path)) {
+      print(paste0("Reading in -> ",data_path))
+      DT <- fread(data_path)
```
terrytangyuan: what if users want to pass additional arguments to fread? I think it's better to just ask users to load dataset themselves, if it's data.table object, then apply an S3 method to do the rest of the processing.

src/interface_r/R/data_prep.R
terrytangyuan: I meant that here you are erroring out when this value is NULL. Why not removing the default value for this argument in function signature so it will error out naturally?

terrytangyuan: Since it seems like this argument is required?

src/interface_r/R/data_prep.R
```diff
@@ -0,0 +1,102 @@
...
+    } else {
+      stop("Parameters `data_table` & `data_path` are NULL. Need to specify at least one.")
+    }
+  } else {
+    DT <- data_table
```
terrytangyuan: Why not making this argument required without default?

src/interface_r/R/data_prep.R
navdeep-G: ah, gotcha. Thats is doable.

src/interface_r/R/data_prep.R
navdeep-G: I will make it required.

src/interface_r/R/data_prep.R
navdeep-G: I will remove the path arg and just make the input `data_table` and it is required.

navdeep-G:
@terrytangyuan Addressed most of your comments

src/interface_r/R/data_prep.R
```diff
@@ -3,10 +3,11 @@
 #'@param data_table `data.table` object containing data that needs to be preprocessed for h2o4gpu
 #'@param response Response column as a string or index
 #'@param save_csv_path Path to save processed data as a csv
+#'@param max_label_encoding_levels The maximum number of uniques required in a column to consider it a categorical variable. Default is 1000
 #'@export
```
terrytangyuan: You can remove "Default is 1000" since users can see the default value in help pages

src/interface_r/R/data_prep.R
```diff
@@ -0,0 +1,85 @@
...
+    if (tt < 2) {
+      print(paste0("Dropping constant column: ", ff))
+      data_table[, (ff):=NULL]
+    }
+  }
+  
+  print(paste0("Number of columns after label encoding: ", ncol(data_table)))
+  
+  numCols <- names(data_table)[which(sapply(data_table, is.numeric))]
+  catCols <- names(data_table)[which(sapply(data_table, is.factor))]
```
terrytangyuan: Consistent variable naming style with others

PR #571, opened 2018-04-21
update packages

pseudotensor:


pseudotensor:
1) for >= vs. ==, we noticed issues with DAI having different versions and actually failed to install something due to the uninstall attempts.  By having == and making DAI and h2o4gpu consistent, no uninstall of higher packages occurs and no problems.

2) For versions chosen, these are latest as of now.  While some will complain that packages are older eventually, we can just update the list manually.  It's dangerous to rely upon >= when things can change to break our build.  I don't like trying to figure out the mysterious reason why jenkins no longer builds.  And usually there is no good reason to upgrade to a new version unless we are encountering a specific problem.  Upgrading arbitrarily is bad idea.

So the only down side is more manual upgrades and users having upgraded versions that we then downgrade.  But if they really want, they can first install h2o4gpu and then upgrade the down-graded packages.  I think that works, right?

PR #578, opened 2018-04-24
Update copyright years in LICENSE

### LINKED ISSUE #577: GitHub cannot infer project license
Based on feedback from @st-pasha:

It seems like the GitHub has problem to figure out right license of the project. There is no license in project header:

image

You can compare it with https://github.com/h2oai/h2o-3 project header:
image

mdymczyk:
Went through the license file in order to fix #577 but not sure why Github isn't recognizing it? Can it be because of the `LICENSES` folder?


navdeep-G:
LGTM

PR #580, opened 2018-04-25
Fail the build on compilation errors.

mdymczyk:
* Fails the build when compilation errors. Until now, even if `cpp` (`.so` compilation) failed, subsequent targets (for example `py`) would run.

* Fails compilation when SWIG cannot find type mappings - this usually happens when we have a typo in the interface file etc. and normally shows only as a warning in SWIG - can cause hard to find bugs in the future.

* Removed `const` from `const char *` because the compiler was saying it's a warning and can cause issues - @navdeep-G do we really want it to be `const`? I can put it back and probably suppress that warning

* Put `-w` which suppresses all warnings durning gcc/nvcc compilation as it was cluttering the logs making it hard to find actual errors. If anyone is actually looking at those warnings all the time (but seeing no one is fixing them I assume no?) and needs them I can remove it.

* Added `llvmlite` in version 0.20.0, looks like we need it as a transitive dep (for scikit?) and the newest version 0.23.0 requires LLVM6 which is not great for us (especially for centos build but was also failing on ubuntu for me).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,7 @@ OPTION(DEV_BUILD  "Dev build" OFF)
 SET(CMAKE_CXX_STANDARD 11)
 SET(CMAKE_CXX_STANDARD_REQUIRED ON)
 SET(CMAKE_POSITION_INDEPENDENT_CODE ON)
+set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -w")
 
 # PythonLibs' PYTHON_INCLUDE_PATH doesn't take into account virtualenv etc.
 # Open to suggestions how to do this better.
@@ -70,7 +71,7 @@ TARGET_LINK_LIBRARIES(cpuh2o4gpu ${BLAS_LIBRARIES})
 #============= BUILD CPU LIBRARY
 
 #============= SWIG
-SET(CMAKE_SWIG_FLAGS "")
+SET(CMAKE_SWIG_FLAGS -Werror)
 #============= SWIG
 
 #============= CPU SWIG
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ OPTION(DEV_BUILD  "Dev build" OFF)
 SET(CMAKE_CXX_STANDARD 11)
 SET(CMAKE_CXX_STANDARD_REQUIRED ON)
 SET(CMAKE_POSITION_INDEPENDENT_CODE ON)
+set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -w")
```
navdeep-G: Minor/Cosmetic: Can make `SET`

requirements_buildonly.txt
```diff
@@ -36,3 +36,4 @@ psutil==5.4.5
 awscli==1.15.6
 # for pycharm
 matplotlib==2.0.2
+llvmlite==0.20.0
```
navdeep-G: Needed as a transitive dep as you mentioned? Was the build failing for you?

src/include/solver/pca.h
```diff
@@ -12,7 +12,7 @@ typedef struct params {
   int X_n;
   int X_m;
   int k;
-  const char *algorithm;
+  char *algorithm;
```
navdeep-G: I think we can change this to `char` with no issues. Also, tests seem to be passing.

src/include/solver/tsvd.h
```diff
@@ -17,7 +17,7 @@ typedef struct params {
   int X_n;
   int X_m;
   int k;
-  const char *algorithm;
+  char *algorithm;
```
navdeep-G: Same as above comment.

navdeep-G:
LGTM

PR #582, opened 2018-04-25
xgboost with GPU glm by Rory

pseudotensor:


Makefile
```diff
@@ -180,7 +180,7 @@ apply-xgboost-cpu-local: libxgboost-cpu-local pipxgboost
 
 pipxgboost:
 	@echo "----- pip install xgboost built locally -----"
-	cd xgboost/python-package/dist && pip install xgboost-0.7-py3-none-any.whl --upgrade --target ../
+	cd xgboost/python-package/dist && pip install *.whl --upgrade --target ../
```
navdeep-G: Will there only be 1 .whl file?

Makefile
```diff
@@ -474,16 +474,16 @@ apply-xgboost-nonccl-cuda9: apply-xgboost-nonccl-local #pipxgboost-nonccl-cuda9
...
 pipxgboost-nccl-cuda9: pipxgboost
 	@echo "----- pip install xgboost-nccl-cuda9 from S3 -----"
-	mkdir -p xgboost/python-package/dist ; cd xgboost/python-package/dist && pip install https://s3.amazonaws.com/artifacts.h2o.ai/releases/bleeding-edge/ai/h2o/xgboost/0.7-nccl-cuda9/xgboost-0.7-py3-none-any.whl --upgrade --target ../
+	mkdir -p xgboost/python-package/dist ; cd xgboost/python-package/dist && pip install https://s3.amazonaws.com/artifacts.h2o.ai/releases/bleeding-edge/ai/h2o/xgboost/0.7-nccl-cuda9/xgboost-0.71-py3-none-any.whl --upgrade --target ../
 pipxgboost-nonccl-cuda9: pipxgboost
 	@echo "----- pip install xgboost-nonccl-cuda9 from S3 -----"
-	mkdir -p xgboost/python-package/dist ; cd xgboost/python-package/dist && pip install https://s3.amazonaws.com/artifacts.h2o.ai/releases/bleeding-edge/ai/h2o/xgboost/0.7-nonccl-cuda9/xgboost-0.7-py3-none-any.whl --upgrade --target ../
+	mkdir -p xgboost/python-package/dist ; cd xgboost/python-package/dist && pip install https://s3.amazonaws.com/artifacts.h2o.ai/releases/bleeding-edge/ai/h2o/xgboost/0.7-nonccl-cuda9/xgboost-0.71-py3-none-any.whl --upgrade --target ../
```
navdeep-G: Going from `.7` -> `.71` is a version bump? Is this what has always been followed? Trying to remember...

pseudotensor: They started doing versions, and they bump every so often.  So instead of tracking the versions, just install whatever we build.

Makefile
pseudotensor: Ya, the directory is cloned freshly.  We can see if jenkins has issues, seems not so far.

Issue #493, opened 2018-03-21
Random Forest regressor producing really bad results


ledell:
There is an issue with the performance of the Random Forest regressor.  I have found that the results are very bad compared to GBM and also to the **randomForest** R package.

```
library(h2o4gpu)
library(reticulate)  # only needed if using a virtual Python environment
use_virtualenv("/home/ledell/venv/h2o4gpu")  # set this to the path of your venv
library(randomForest)

# Load a sample dataset for regression
# Source: https://archive.ics.uci.edu/ml/datasets/Combined+Cycle+Power+Plant
df <- read.csv("https://github.com/h2oai/h2o-tutorials/raw/master/h2o-world-2017/automl/data/powerplant_output.csv")

# Randomly sample 80% of the rows for the training set
set.seed(1)
train_idx <- sample(1:nrow(df), 0.8*nrow(df))

# Create train & test; last (5th) column is the response
x_train <- df[train_idx,-5]
y_train <- df[train_idx,5]
x_test <- df[-train_idx,-5]
y_test <- df[-train_idx,5]

# Train three models
model_gbr <- h2o4gpu.gradient_boosting_regressor() %>% fit(x_train, y_train)
model_rfr <- h2o4gpu.random_forest_regressor() %>% fit(x_train, y_train)
rrf <- randomForest(x = x_train, y = y_train)

# Generate predictions
pred_gbr <- model_gbr %>% predict(x_test)
pred_rfr <- model_rfr %>% predict(x_test)
pred_rrf <- predict(rrf, x_test)

# Compare test set performance using MSE
mse(actual = y_test, predicted = pred_gbr)        # 15.2938
mse(actual = y_test, predicted = pred_rfr)    # 25168.17  #YIKES!
mse(actual = y_test, predicted = pred_rrf)         # 10.58695
```

Another dataset:
```
df <- read.csv("https://archive.ics.uci.edu/ml/machine-learning-databases/abalone/abalone.data", header = FALSE)
df[,1] <- as.integer(df[,1])

# Randomly sample 80% of the rows for the training set
set.seed(1)
train_idx <- sample(1:nrow(df), 0.8*nrow(df))

# Create train & test; last (9th) column is the response
x_train <- df[train_idx,-9]
y_train <- df[train_idx,9]
x_test <- df[-train_idx,-9]
y_test <- df[-train_idx,9]

# Train three models
[...]

PR #583, opened 2018-04-25
XGBoost wrappers parameters missing

### LINKED ISSUE #336: XGBoost wrappers parameters missing
http://xgboost.readthedocs.io/en/latest/python/python_api.html#xgboost.XGBClassifier and the Regressor have parameters which we don't expose in our wrappers, we probably should. Also, we probably need to pass somehow kwargs.

navdeep-G:
* Resolves #336 by adding full XGBoost params for GBM impl along with `**kwarg`argument.
* Resolves https://github.com/h2oai/h2o4gpu/issues/493 by setting `n_estimators` to 100 by default in RF impl.


navdeep-G:
TODO: Update R wrappers.

navdeep-G:
@terrytangyuan I have updated it to account for `**kwargs` in the R API.

navdeep-G:
@terrytangyuan Adjusted the R API to follow the Google R style guide.

@mdymczyk Yes, but I think we can add more on a case by case basis. I do not fully understand XGBoost in terms of Random Forest, but the impl is more "Random Forest like" in a sense. The main parameter I can see guiding a Random Forest type impl is `num_parallel_tree`,which is described here:
```
num_parallel_tree Experimental parameter. number of trees to grow per
round. Useful to test Random Forest through Xgboost (set colsample_bytree < 1,
subsample < 1 and round = 1) accordingly. Default is 1.
```
Also, here is a R code snippet from XGBoost's R documentation that describes how to run a RF like model:
```
## random forest-like model of 25 trees for binary classification:
set.seed(11)
bst <- xgboost(data = train$data, label = train$label, max_depth = 5,
nthread = 2, nrounds = 1, objective = "binary:logistic",
num_parallel_tree = 25, subsample = 0.6, colsample_bytree = 0.1)
```

PR #592, opened 2018-05-04
Add function to find optimal n_components for decomposition methods

### LINKED ISSUE #589: Add function to find optimal k/n_components for TSVD/PCA

navdeep-G:
* Resolves #589


RAMitchell:
I think this can be implemented in a more pythonic way by composition of the two functions np.cumsum and bisect.

A couple of questions relating to the scope of h2o4gpu.

What is the purpose of this function? It seems like something a user could trivially do in python and I don't see an sklearn analogue. Are we planning to support other utility functions? If so what is the criteria for inclusion in h2o4gpu?

PR #595, opened 2018-05-07
Factorization

stjoern:
Acc. matrix factorization

(1) as input for parsing data (numpy, file input), Eigen library part used
(2) computation used https://devblogs.nvidia.com/parallelforall/accelerate-recommender-systems-with-gpus/
(3) output, the same Eigen library with api to numpy.


stjoern:
@mdymczyk 
the matrix factorization is to be found in src/mxfactorization
For build, just make, will be for testing purposes main executable created. 
As input data, one needs testing and training data sets, can be downloaded from here:
[https://s3.amazonaws.com/matrix-factorization-data/reverse_netflix.mm?versionId=OCrxX6BgtKq4xRddTS7uzAS9t8ZvSa0p](url) - training data
[https://s3.amazonaws.com/matrix-factorization-data/reverse_netflix.mme?versionId=7xFTMZhbYahjx.pjTtuvo0w4IJ51mYni](url) - testing data.
The output is as float vectors of thetas.
I will be happy for your help with swig and embedding FM into system.

mdymczyk:
@stjoern 

For the binary generation step, we'll need to get rid of that Makefile in matrixfactorization. For now you can remove all the handling of most architectures, we just need to be sure it builds on linux (x86_64 and ppc64le). The flags we are using in CMakeLists should be enough so probably don't need to change anything.

We want to start with building 2 object files, one with only CPU code, one with CPU+GPU code. You can see how it's done for example here: https://github.com/h2oai/h2o4gpu/blob/master/CMakeLists.txt#L44 it says which files should be assigned to `COMMON_SOURCES`, then you make it into an object file with `ADD_LIBRARY(commonh2o4gpu OBJECT ${COMMON_SOURCES})` and you can link it to our resulting `.so` by doing `ADD_LIBRARY(cpuh2o4gpu STATIC ${CPU_SOURCES} $<TARGET_OBJECTS:commonh2o4gpu>)`, so you will want something like `cpumatrixfactorization` (with only CPU code) and then later in that file a similar `gpumatrixfactorization` linked to `gpuh2o4gpu`.

Swig files are here https://github.com/h2oai/h2o4gpu/tree/master/src/swig

You will need to make an interface file inside the `solver` folder (please refer to the existing ones, our example https://github.com/h2oai/h2o4gpu/blob/master/EXAMPLE_SOLVER.md and SWIG doc)

You will need to also include that file in both https://github.com/h2oai/h2o4gpu/blob/master/src/swig/ch2o4gpu_cpu.i and https://github.com/h2oai/h2o4gpu/blob/master/src/swig/ch2o4gpu_gpu.i

Finally we will need some Python wrappers, does scikit learn have matrix factorization or will we have to mimic some other library?

stjoern:
@mdymczyk 

It's implemented and I tried to make the build based on your instruction. Please take a look and tell me what's needed further.
In terms of CPU version, unfortunately I've implemented only GPU version, no CPU version is now available.  I would need probably make some python tests or something where I could call the one method using the algorithm, but please first take a look if it is correct now as it is.

PR #596, opened 2018-05-08
Ensure namespaces are organized and named properly

navdeep-G:
* Changed `Matrix` namespace from `tsvd` to `matrix`
* Changed DeviceContext namespace to `device`
* Moved `util.cu` from `tsvd/` to `src/gpu/utils` and changed namespace to `h2o4gpu`
* Removed use of `typedef` in Matrix class and opted to just use `float` and `double`

### DIFF
--- a/src/gpu/data/matrix.cu
+++ b/src/gpu/data/matrix.cu
@@ -5,6 +5,9 @@
 
 namespace tsvd
 {
+	using namespace device;
+	using namespace h2o4gpu;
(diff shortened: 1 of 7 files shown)


mdymczyk:
LGTM

navdeep-G:
Yes, that’s odd. I do see this branch is one commit behind master though. But that shouldn’t restart the build right?

PR #618, opened 2018-05-26
Fail when NVML not found, look also for stubs

mdymczyk:
Certain docker images don't have the actual libnvidia-ml.so, only stubs. These stubs (from what I can tell) are under `/usr/local/cuda-X.Y`. Stubs seem to be enough during build time as long as the actual binary is on the system during runtime.

Also added a failure mechanism if NVML is not present during build time, otherwise the build would pass but the runtime would fail (not happening in master branch because we're not testing PPC64 but is failing in the centos rebuild branch).


mdymczyk:
Hm looks like some nodes are still failing - need to check if they are misconfigured or I PRd a too broad solution (saw 384 drivers somewhere where only 396 should be).

Issue #602, opened 2018-05-09
Host R package documentation


terrytangyuan:
We now have R package documentation here: https://github.com/h2oai/h2o4gpu/tree/master/src/interface_r/docs

@angela0xdata Could you help us host it in a similar way to what's done for h2o-3 R API docs? Or tag anyone who's able to help with this.

cc: @ledell @navdeep-G

angela0xdata:
Everything looks good locally. I'll work with Anmol to get these docs included in the build.

terrytangyuan:
Thanks!

abal5:
I'll put this into a Jenkinsfile stage so we can get it building.

terrytangyuan:
@abal5 Could you point me to the documentation page? Perhaps add the link to R package's README?

angela0xdata:
@terrytangyuan  http://h2o-release.s3.amazonaws.com/h2o4gpu/nightly/h2o4gpu-Rdocs/7/interface_r/docs/index.html
(Note an issue still exists with the top heading... and we still have no idea why it's getting cut off.)

terrytangyuan:
Does it happen locally when you open the html? Also probably try previous versions of pkgdown.

angela0xdata:
@Erin, no Anmol has been working on getting h2o4gpu R docs into the build and then uploaded to S3. 

Here is the S3 link:
http://h2o-release.s3.amazonaws.com/h2o4gpu/nightly/h2o4gpu-Rdocs/7/interface_r/docs/index.html

Also attaching a zip of the files when built locally:
[docs.zip](https://github.com/h2oai/h2o4gpu/files/2206364/docs.zip)

Regarding latest stable vs nightly, I don't know of a latest stable yet. I don't believe there have been any H2O4GPU releases since this issue was closed. (But I could be wrong.)

The docs will reside in the H2O4GPU box on docs.h2o.ai, just like other products. Currently, that box only has links to the README, the Apache license, and a "download" link pointing to github. I can add it to the R box also.

PR #619, opened 2018-05-30
Generate R docs

abal5:
#602 

To add the generation of R documentation for this project.
Using a small container from dockerhub that has R 3.1 on it, install the packages needed, build the docs, archive the doc files.


abal5:
I rather have this build separately as well so I'll look into making this it's own jenkinsfile.

navdeep-G:
Stage looks good. I agree it should be built separately.

Jenkinsfile.utils2
```diff
@@ -199,6 +199,23 @@ pipeline {
...
+            steps {
+                sh """
+                    R CMD INSTALL usethis
+                    R CMD INSTALL devtools
+                    R CMD INSTALL  pandoc
+                    wget https://github.com/r-lib/pkgdown/archive/v1.0.0.tar.gz
+                    R CMD INSTALL pkgdown-1.0.0.tar.gz
+                    R && `pkgdown::build_site(examples=FALSE)` && q(save="no")
+                   """
+                archiveArtifacts artifacts: 'src/interface_r/docs/*.html'
```
terrytangyuan: This doesn’t include images if there’s any? Also there’s a index.html in a directory above that you might need as well

abal5:
Ok so the R documentation is now being generated and being uploaded to S3, the only thing left is we need to decide what kind of url/folder path we would like and also how versioning should be handled.

PR #627, opened 2018-06-19
Pass verbosity to c get info method

mdymczyk:
Some users are getting segfaults in this method (bitfusion) - hopefully more logs will help to pinpoint the problem.

### DIFF
--- a/src/gpu/utils.cu
+++ b/src/gpu/utils.cu
@@ -11,6 +11,8 @@
 #include <iostream>
 #include <assert.h>
 
+#include "../common/logger.h"
+
 #include <cuda_runtime_api.h>
 #include "nvml.h"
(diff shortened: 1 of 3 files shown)


pseudotensor:
Sure, but that's not how it's used right now.  There are no other levels except 0 and 1, so I prefer to keep it as False and True for now so DAI and other uses do not break.

mdymczyk:
@pseudotensor how about this, the API doesn't break (False is automatically handled as 0 and True prints everything) but also supports finer grained logging.

pseudotensor:
Can you add check if verbose == False and set verbose = 0 too?

mdymczyk:
@pseudotensor sure, done.

PR #629, opened 2018-06-22
Remove assertions.

mdymczyk:
Don't use assertions as they are compiled out with mode release in cmake, instead use status code.

Should we return `None` in Python (as I'm doing now), throw an exception or something else?


pseudotensor:
ya, don't return rv

mdymczyk:
@pseudotensor makes sense, done

PR #647, opened 2018-07-16
Add versioning to numpy and other things so consistent with requireme…

pseudotensor:


pseudotensor:
Could automate, but for now need to make consistent.

PR #662, opened 2018-07-30
Fixes for glm notebook, broken parallel run for glm, and an xgboost/src build issue

pseudotensor:
* Fixes glm notebook, splitting animation and run due to unknown issue.
* minor fixes for build
* Fix broken parallel mode for glm


pseudotensor:
Hi @mdymczyk , when you converted to cmake/centos, the USEPARALLEL build parameter was lost, so glm was always using 1 GPU and 1 thread for OpenMP.  I know @tomkraljevic had problems with this for power and/or centos, but this loses a major functionality.

I changed things so x86 works in ubuntu  now, but centos complains about pragma's and openmp:

pseudotensor:
https://www.ideaconsult.net/web/luchesar/blogs/-/blogs/imagemagick-fails-with-error:-expected-pragma-omp-clause-before-omp_throttle

Not sure if related.  We do use ccache.

pseudotensor:
Also, the current h2o4gpu makes completely wrong predictions and actuals for that mapd notebook.  It should be income earned on order 20000+ but actuals are order 0.2-0.8 and predictions are all over the place within 0-1.0.

So probably the gpu pointer memory stuff not working.  Will see if older h2o4gpu works or make better unit test to compare non-ptr to ptr method results.

mdymczyk:
Regarding:

```
/root/repo/src/common/elastic_net_ptr.cpp:1507:22: error: expected ‘#pragma omp’ clause before ‘proc_bind’

 #pragma omp parallel proc_bind(master)
```

This is because `proc_bind` requires openmp >=4.0 (or 4.0.1 don't remember) which comes bundled with gcc >= 4.9 and I think the one we use in the docker is 4.8.x. We do install `libgomp` explicitly but not sure which version it installs on power machines. Logs for XGBoost show that its cmakelists finds `18:27:30 -- Found OpenMP: TRUE (found version "3.1")` 3.1 by default. 

So we need to check if it's because there are multiple versions on that system (original one + one we installed) or whether it's because `yum install libgomp` installs an old version and we need to compile a newer one by hand.

Maybe setting a required version in our CMakeLists will be enough.

PR #686, opened 2018-09-27
Re-enable power tests

pseudotensor:


abal5:
"ibm-power-gpu" is the only machine right now and also keep in mind this is shared by IBM with us, we already lost 1 machine a few weeks ago "ibm-power-gpu2" we have a power9 box in house which is mr-0xp9

sh1ng:
Not relevant any more.

Issue #680, opened 2018-09-19
Issue getting a working python notebook in Jupyter docker images


hemenkapadia:
Following the instructions in section [Running Jupyter Notebooks with Docker](https://github.com/h2oai/h2o4gpu#running-jupyter-notebooks-with-docker) of the [README](https://github.com/h2oai/h2o4gpu) file results in the below kernel error when trying to launch a Python 3 notebook from the Jupyter interface.

```
Traceback (most recent call last):
  File "/opt/h2oai/h2o4gpu/python/lib/python3.6/site-packages/traitlets/traitlets.py", line 528, in get
    value = obj._trait_values[self.name]
KeyError: 'loop'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/h2oai/h2o4gpu/python/lib/python3.6/site-packages/tornado/web.py", line 1592, in _execute
    result = yield result
  File "/opt/h2oai/h2o4gpu/python/lib/python3.6/site-packages/tornado/gen.py", line 1133, in run
    value = future.result()
  File "/opt/h2oai/h2o4gpu/python/lib/python3.6/site-packages/tornado/gen.py", line 1141, in run
    yielded = self.gen.throw(*exc_info)
  File "/opt/h2oai/h2o4gpu/python/lib/python3.6/site-packages/notebook/services/sessions/handlers.py", line 73, in post
    type=mtype))
  File "/opt/h2oai/h2o4gpu/python/lib/python3.6/site-packages/tornado/gen.py", line 1133, in run
    value = future.result()
  File "/opt/h2oai/h2o4gpu/python/lib/python3.6/site-packages/tornado/gen.py", line 1141, in run
    yielded = self.gen.throw(*exc_info)
  File "/opt/h2oai/h2o4gpu/python/lib/python3.6/site-packages/notebook/services/sessions/sessionmanager.py", line 79, in create_session
    kernel_id = yield self.start_kernel_for_session(session_id, path, name, type, kernel_name)
  File "/opt/h2oai/h2o4gpu/python/lib/python3.6/site-packages/tornado/gen.py", line 1133, in run
    value = future.result()
  File "/opt/h2oai/h2o4gpu/python/lib/python3.6/site-packages/tornado/gen.py", line 1141, in run
    yielded = self.gen.throw(*exc_info)
[...]

PR #690, opened 2018-10-01
PR to make docker rumtime work for demos #680

hemenkapadia:
PR to fix issue #680. Aim to get docker runtime working for H2O4GPU demos. Should also fix issue with docker files being created with very small size due to 
```
Error response from daemon: file integrity checksum failed for "usr/lib64/libdb-5.3.so"
```
This is observed for CUDA 9 and 9.2. CUDA 8 seems fime


hemenkapadia:
Closed. Do not Merge. See PR#691 instead

PR #691, opened 2018-10-02
Another approach to solve issue #680

hemenkapadia:
Forcing torando to version known to work


hemenkapadia:
@pseudotensor , updated the README to highlight default docker settings is to not allow remote access to Jupyter + instructions on how to get it enabled.

Merge post review.

PR #705, opened 2019-01-11
fix build

sh1ng:
* use pip directly
* use python from configuration
* include numeric to support std::accumulate


mdymczyk:
@sh1ng sorry for the late review! This looks good with one minor concern - not sure we should push any changes to the `xgboost` submodule, since it won't be reflected in the actual github repo and if someone updates it we'll loose the changes (iirc).

It's our private fork of xgboost so either myself or @pseudotensor (if he has no objections) can do it and update the submodule here.

PR #706, opened 2019-01-17
Fix typo in truncated svd algorithm

pseudotensor:


pseudotensor:
leads to failure in DAI

PR #709, opened 2019-01-22
fix CI & CUDA forward compatibility

sh1ng:
The following issues are fixed:
- forward compatibility, there was an issue on CUDA 8 + Tesla V at least
- https://github.com/h2oai/h2o4gpu/issues/711
- CI: triggers, check previous state, etc  
- adjusted tolerance when comparing with _sklearn_
- disabled tests for _lgbm_ on _ppc64le_
- using a single GPU for _xgboost_ on _ppc64le_
- updated _Apache arrow_ to 0.12.0
- Docker-configuration for _ppc64le_
- using the same version of _numpy_ everywhere to speed-up CI
- increased number of _pytest_ workers

Are going to be fixed in separate PRs:
- lgbm with gpu on ppc64le
- xgbbost with NCCL on ppc64le


sh1ng:
On a machine with V100 and CUDA 8 there's an error 
```
terminate called after throwing an instance of 'thrust::system::system_error'
  what():  function_attributes(): after cudaFuncGetAttributes: invalid device function
Aborted
```

Dockerfile-build
```diff
@@ -155,7 +155,7 @@ RUN bash -c 'if [ `arch` = "ppc64le" ]; then \
 #
 # Install Python requirements
 #
-RUN pip install numpy==1.14.5 scipy==1.1.0 setuptools==39.0.1
+RUN pip install numpy==1.15.1 scipy==1.1.0 setuptools==39.0.1
```
pseudotensor: Was there a need to update numpy?  Just want to avoid mismatch with DAI if possible for now till we figure out.

Dockerfile-runtime
```diff
@@ -41,28 +41,17 @@ WORKDIR $HOME
...
-ENV PILLOW_VERSION=4.2.1
-RUN bash -c 'if [ `arch` = "ppc64le" ]; then \
-	wget https://files.pythonhosted.org/packages/55/aa/f7f983fb72710a9daa4b3374b7c160091d3f94f5c09221f9336ade9027f3/Pillow-${PILLOW_VERSION}.tar.gz && \
-	tar xvf Pillow-${PILLOW_VERSION}.tar.gz && \
-	cd $HOME/Pillow-${PILLOW_VERSION} && \
-	sed -i "s/'ppc64'/'ppc64le'/g" setup.py && \
-	python setup.py install && \
-	cd $HOME && \
-	rm -rf Pillow-${PILLOW_VERSION}*; \
+	python setup.py install; \
```
pseudotensor: So pillow wasn't needed after your fixes?  That's great.

Makefile
```diff
@@ -416,7 +416,7 @@ dotest:
 	rm -rf ./tmp/
 	mkdir -p ./tmp/
   # can't do -n auto due to limits on GPU memory
-	pytest -s --verbose --durations=10 -n 3 --fulltrace --full-trace --junit-xml=build/test-reports/h2o4gpu-test.xml tests/python/open_data 2> ./tmp/h2o4gpu-test.$(LOGEXT).log
+	pytest -s --verbose --durations=10 --numprocesses 6 --fulltrace --full-trace --junit-xml=build/test-reports/h2o4gpu-test.xml tests/python/open_data 2> ./tmp/h2o4gpu-test.$(LOGEXT).log
```
pseudotensor: Caution, can run out of GPU memory easily.

TROUBLESHOOTING.md
```diff
@@ -18,7 +18,7 @@ version we tried to install.  E.g. on conda you might need to do:
 
 ```
 pip uninstall numpy
-pip install numpy==1.14.5 # or whatever version was attempted to be installed by the wheel
+pip install numpy==1.15.1 # or whatever version was attempted to be installed by the wheel
```
pseudotensor: Same here, needed?

ci/Jenkinsfile-ppc64le-cuda8
```diff
@@ -28,7 +28,7 @@ def labelruntime = "ibm-power || ibm-power-gpu"
 def doingbenchmark = "0"
 def dobenchmark = "0"
 def doruntime = "1"
-def data_dirs = "-v /home/jenkins/smalldata:/smalldata -v /home/jenkins/open_data:/open_data"
+def data_dirs = "-v /home/0xdiag/h2o4gpu/data:/data -v /home/0xdiag/h2o4gpu/open_data:/open_data"
```
pseudotensor: Great, so jenkins ppc setup was just wrong for the data path?  Same path as x86?

ci/Jenkinsfile-ppc64le-cuda8
```diff
@@ -158,7 +158,7 @@ pipeline {
                         echo "After Stashed wheel file:"
                         ls -l src/interface_py/dist/${platform}/  || true
                        """
-//                     //h2o4gpuUtils.runTests(buildInfo.get(), "${cuda}", "${extratag}", "${platform}", "${testtype}", "${data_dirs}")
+                    h2o4gpuUtils.runTests(buildInfo.get(), "${cuda}", "${extratag}", "${platform}", "${testtype}", "${data_dirs}")
```
pseudotensor: Great

ci/Jenkinsfile.template
```diff
@@ -195,11 +195,11 @@ pipeline {
                     }
                 }
                 }
-                retryWithTimeout(1000 /* seconds */, 5 /* retries */) {
-                    script {
-                        h2o4gpuUtils.publishToPoboys("${platform}")
-                    }
-                }
+                //retryWithTimeout(1000 /* seconds */, 5 /* retries */) {
```
pseudotensor: why not needed?  I thought required for conda?  @hemenkapadia

tests/python/open_data/gbm/test_xgb_sklearn_wrapper.py
```diff
@@ -2,12 +2,17 @@
...
 import logging
 import pandas as pd
+import platform
 
 print(sys.path)
 
 
 logging.basicConfig(level=logging.DEBUG)
 
+# TODO: remove when support NCCL on ppc64le platform
```
pseudotensor: nccl supported on cuda9.2 or cuda10 for ppc I think already.  Dealt with that before, we can chat if not clear.

Dockerfile-build
hemenkapadia: I also recollect that there were some issues with complete conda package dependency tree when numpy was increased. Though that may have changed. 

our build process should ensure that both pip and conda packages can be created but i believe that would have been taken care. Just adding note as caution.

ci/Jenkinsfile-ppc64le-cuda9
```diff
@@ -206,6 +206,16 @@ pipeline {
...
+                    sh """
+                        echo "Before unstash condapkg:"
+                        ls -l condapkgs  || true
+                        rm -rf condapkgs || true
+                       """
+                    unstash "condapkg"
+                    sh """
+                        echo "After unstash condapkg:"
+                        ls -l condapkgs  || true
+                       """
```
hemenkapadia: I was never able to build conda package for h2o4gpu for ppc64le, hence the corresponding Jenkins files did not have any conda specific code. Are we now able to build it for ppc64le too? That would be awesome step forward for cross arch scoring pipeline deployment

ci/Jenkinsfile-ppc64le-cuda92
```diff
@@ -206,6 +206,16 @@ pipeline {
...
+                    sh """
+                        echo "Before unstash condapkg:"
+                        ls -l condapkgs  || true
+                        rm -rf condapkgs || true
+                       """
+                    unstash "condapkg"
+                    sh """
+                        echo "After unstash condapkg:"
+                        ls -l condapkgs  || true
+                       """
```
hemenkapadia: same as above

ci/Jenkinsfile.template
hemenkapadia: Yes it is needed IMHO. We have an internal anaconda server where we upload conda packages for our builds. http://poboys.h2o.ai:6969/poboys/pkgs/linux-64  contains h2o4gpu builds. It allows one click relase to anaconda.org once we finalize a build version to release.

sh1ng: Looks this part was removed manually from every config file, so I've decided to remove(comment out actually) it from the template. I'll add it back and regenerate configurations.

Dockerfile-build
sh1ng: There's a discrepancy between version in Docker file and https://github.com/h2oai/h2o4gpu/blob/master/src/interface_py/requirements_runtime.txt. It results in numpy=1.14 get installed and uninstalled to install version 1.15. Using the same version can speed-up CI a bit.

Dockerfile-runtime
sh1ng: I haven't found any explicit dependency on it. Plus all tests are passed.

Makefile
sh1ng: It should not be a problem for CI(never seen error related to memory allocation). For local testing we would had to reduce it, but my main focus is CI.

TROUBLESHOOTING.md
sh1ng: I think so

ci/Jenkinsfile-ppc64le-cuda8
sh1ng: Yes and Yes.

Also sync jobs were not running.

TROUBLESHOOTING.md
pseudotensor: Yes, ok, DAI has 1.15.1, so is good change.

PR #721, opened 2019-02-06
nccl as a submodule, xgboost live-lock fix

sh1ng:
* Updated NCCL to _2.4.2_
* No need to install NCCL separately
* It works on pcc64le with CUDA8/9 
* ~~xgboost's live-lock still occurs(I hoped it fixed in a new version). Looks like it's xgboost issue https://github.com/dmlc/xgboost/issues/4107~~ Merged a temporary fix. Deadlock can still occur on CUDA 8(know issue of NCCL)
* Added workaround(cache build in docker cache) to not compile NCCL every time. Another option could be to store it in s3, but that introduces a dependency on H2O infrastructure.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -18,4 +18,4 @@
 	url = https://github.com/h2oai/LightGBM.git
 [submodule "nccl"]
 	path = nccl
-	url = git@github.com:NVIDIA/nccl.git
+	url = https://github.com/NVIDIA/nccl.git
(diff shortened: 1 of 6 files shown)


trivialfis:
@sh1ng Latest NCCL doesn't solve the problem, I just tried to compile NCCL from source.  I will workaround the problem by using old `Group*` APIs of NCCL in XGBoost.

PR #726, opened 2019-02-22
lightgbm on ppc64le

sh1ng:


sh1ng:
@pseudotensor FYI

Disabling altivec instruction( `-mno-altivec`) allows to compile on ppc64le platform. thumb up to https://github.com/Microsoft/LightGBM/issues/1905  

Unfortunately Nvidia doesn't support OpenCL on power platform.  
https://devtalk.nvidia.com/default/topic/1025683/cuda-setup-and-installation/opencl-on-power8/

PR #729, opened 2019-03-05
Matrix factorization

sh1ng:
* Based on https://github.com/cuMF/cumf_als
* GPU only version. Do we need implement CPU version?
* The code is quite awkward, I've started to make it more structured, but not finished
* Only support float32 for now
* Only single GPU for now 
* A few more MUST TO HAVE features(like releasing memory, more granular python API) are in progress and will be in next commits
* There're a few copies of train dataset on GPU(CSC, CSR, COO sparse matrices), I'm going to fix it in a separate PR

But review can be started.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -179,4 +179,4 @@ src/interface_py/build_info.txt
 
 # Temp requirement file generated for Conda
 src/interface_py/requirements_conda.txt
-open_data/
+open_data/*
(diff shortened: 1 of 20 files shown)


sh1ng:
The implementation is unstable when use LU solver instead _CG_
```
#define USE_CG // comment it out  
```
But as CG is considered as a better alternative than "exact" solution, it's ok for now.

PR #751, opened 2019-04-04
build lightgbm with gcc

sh1ng:


tomkraljevic:
I did a test by slamming in the h2ogpu built from the PR inside DAI 1.6.0 on Power and it worked for me.

PR #757, opened 2019-04-11
[NOT FOR MERGING] Build lgbm with gcc on power, a few more tests (#754)  -- from df03261

pseudotensor:


pseudotensor:
@sh1ng This isn't working as-is, other changes you've made mean it won't build like this.  FYI, I started like:

```
git show df032619591eefb531c74952529a6124b0ded2a5 > diff.txt
git checkout -b rel-dai-1.6.1 36dac53 -f
git apply diff.txt
git commit -a -m "Build lgbm with gcc on power, a few more tests (#754)  -- from df03261"
git push origin rel-dai-1.6.1
```

PR #761, opened 2019-04-15
build on cuda-10

sh1ng:


sh1ng:
Yes, otherwise we'd not see `*-cida10` checks

sh1ng:
hm.. it's weird 

Screenshot from 2019-04-24 13-09-01

I'm going to update Readme after the PR is merged and we have bleeding-edge version for cuda-10. 
Verified.

PR #768, opened 2019-05-24
fix build on CUDA 10.1:

sh1ng:
* use __shfl_down_sync
* add kmeans copy overload methods for device_allocator
* linking with `libcublasLt_static.a`

Besides the fact that we're not going to support CUDA-10.1 for now(see https://github.com/dmlc/xgboost/issues/4264). It's better to be compatible with it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,13 +105,33 @@ if(USE_CUDA)
                 -DHAVECUDA
         )
 
+        find_library(CUBLAS_STATIC_LIB NAMES libcublas_static.a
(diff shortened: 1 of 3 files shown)


sh1ng:
It's not linking on CUDA-10.1 because NVidia made extra changes on libcublas.
https://devtalk.nvidia.com/default/topic/1048658/gpu-accelerated-libraries/linker-problem-with-cuda-10-1-105-on-cento-7-undefined-references/
https://github.com/tensorflow/tensorflow/issues/26150

PR #770, opened 2019-05-27
Allow lgbm to fully pickle

pseudotensor:


pseudotensor:
I removed my fix and used microsoft's more complete and tested fix.

PR #772, opened 2019-05-30
add a few xgboost tests varying num_gpus

sh1ng:


RAMitchell:
Feel free to migrate any tests to us. If we have the tests then it prevents us committing anything that breaks your functionality in the first place.

PR #775, opened 2019-05-31
ppc hang

sh1ng:


sh1ng:
While it doesn't solve NCCL hang on ppc it contains a few useful changes that I'd like to have.

PR #777, opened 2019-06-07
enable NCCL

sh1ng:


pseudotensor:
https://github.com/h2oai/xgboost/commits/h2oai

not sure what is done here, but usually we update the xgboost h2oai branch since we use that in h2o4gpu.  That's what I did above.

Can you explain what you are doing here?

PR #786, opened 2019-07-17
Remove unnecessary dependencies

sh1ng:


sh1ng:
@pseudotensor do you have any idea why `requirements_runtime.txt` was added?

Dockerfile-runtime
```diff
@@ -64,6 +64,7 @@ RUN \
   pip install --no-cache-dir -r requirements.txt && \
   pip install --no-cache-dir -r requirements_runtime_demos.txt
 
+RUN pip install --no-cache-dir pytest==3.10.1 pytest-forked==0.2 pytest-xdist==1.22.2 pytest-cov==2.5.1 pylint==1.8.4 pytest-timeout==1.3.3
```
mmalohlava: In dai we have these requirements in a separated file: `requirements-test.txt` and it is install in docker in a normal way (`pip install -r ...`)

pseudotensor: Ya, I'd prefer if we only had super basic (pip and setuptools) in Dockerfile, and rest as requirements.

Dockerfile-build
```diff
@@ -175,12 +175,8 @@ RUN bash -c 'if [ `arch` = "ppc64le" ]; then \
 RUN pip install numpy==1.16.4 scipy==1.2.1 setuptools==39.0.1
```
pseudotensor: These are here only because they need to be done super early to avoid issue when doing full requirements file handling by pip.  But nominally, let's do everything in requirements except for these basic things.

mmalohlava:
@pseudotensor can we merge it?

PR #797, opened 2019-09-24
ARIMA n-step regression

sh1ng:
Implementation of ARIMA model. It contains CUDA part only and a few more changes are upcoming, but I believe it's good enough to be reviewed. After python part is done I'm going to add a few more methods to evaluate coefficients or a c++ version without CUDA.  

Some theory can be found in:
* http://www.phdeconomics.sssup.it/documents/Lesson12.pdf
* https://math.unice.fr/~frapetti/CorsoP/Chapitre_4_IMEA_1.pdf


sh1ng:
There's instability in convergence probably due to residual approximation. I think I can be fixed by implementing IIR filter(residual computation can be expressed as an IIR filter).

Most suitable approach that I've found https://userweb.cs.txstate.edu/~burtscher/papers/asplos18.pdf

PR #820, opened 2020-01-20
skip RPATH for lightgbm

sh1ng:


Makefile
```diff
@@ -119,8 +119,8 @@ lightgbm_gpu:
 	cd LightGBM && (rm -rf build || true) && mkdir -p build && \
 	sed -i 's/#define BOOST_COMPUTE_USE_OFFLINE_CACHE//g' src/treelearner/gpu_tree_learner.h && \
 	cd build && \
-	cmake .. -DUSE_GPU=1 -DCMAKE_C_COMPILER=/usr/bin/gcc -DCMAKE_CXX_COMPILER=/usr/bin/g++ -DOpenCL_LIBRARY=$(CUDA_HOME)/lib64/libOpenCL.so -DOpenCL_INCLUDE_DIR=$(CUDA_HOME)/include/ -DBOOST_ROOT=/opt/boost -DBoost_USE_STATIC_LIBS=ON -DBoost_NO_SYSTEM_PATHS=ON && \
-	make OPENCL_HEADERS=$(CUDA_HOME)/targets/x86_64-linux/include LIBOPENCL=$(CUDA_HOME)/targets/x86_64-linux/lib -j && cd .. && \
+	cmake -DUSE_GPU=1 -DCMAKE_SKIP_RPATH=1 -DCMAKE_C_COMPILER=/usr/bin/gcc -DCMAKE_CXX_COMPILER=/usr/bin/g++ -DOpenCL_LIBRARY=$(CUDA_HOME)/lib64/libOpenCL.so -DOpenCL_INCLUDE_DIR=$(CUDA_HOME)/include/ -DBOOST_ROOT=/opt/boost -DBoost_USE_STATIC_LIBS=ON -DBoost_NO_SYSTEM_PATHS=ON .. && \
+	make -j && cd .. && \
```
pseudotensor: Those other make options were required at the time.

sh1ng: May be we no need it any more. There's no information about it in https://lightgbm.readthedocs.io/en/latest/GPU-Tutorial.html

sh1ng:
We could use 
`make install` instead of setting `CMAKE_SKIP_RPATH` and publish lib.so into local folder(cmake will remove RPATH), but it doesn't work.

pseudotensor:
Ok, as long as we have some lgbm tests that pass it's good.

PR #821, opened 2020-01-24
Update xgboost (had to fixup conflicts) and lightgbm (no conflicts)

pseudotensor:

PR #822, opened 2020-02-04
xgboost lightgbm update, add xgboost 0.90 and migration script 

sh1ng:
- Branched off #821 but decided to create a separate PR
- To make xgboost 0.90 and 1.0 compatible I had to update dmlc core to use the same version
- Tests are still failing because of https://github.com/dmlc/xgboost/issues/5276
-  If you know better way to handle xgboost exception let me know
- All names are discussable

### DIFF
--- a/src/interface_py/h2o4gpu/util/xgboost_migration.py
+++ b/src/interface_py/h2o4gpu/util/xgboost_migration.py
@@ -1,0 +1,34 @@
+def load_pkl(name):
+    """Load xgboost model from pickle and perform conversion from version 0.90 if require.
+
+    :return:
+        Total number of GPUs and total available memory
+    """
+    import pickle
+    import xgboost
+    with open(name, 'rb') as f:
+        try:
+            model = pickle.load(f)
+            return model
+        except xgboost.core.XGBoostError as e:
+            if "Check failed: header == serialisation_header_" in str(e):
+                import xgboost_prev
+                import tempfile
+
+                class Unpickler(pickle.Unpickler):
+                    def find_class(self, module, name):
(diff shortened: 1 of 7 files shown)


src/interface_py/h2o4gpu/util/xgboost_migration.py
```diff
@@ -0,0 +1,34 @@
...
+    with open(name, 'rb') as f:
+        try:
+            model = pickle.load(f)
+            return model
+        except xgboost.core.XGBoostError as e:
+            if "Check failed: header == serialisation_header_" in str(e):
+                import xgboost_prev
+                import tempfile
+
+                class Unpickler(pickle.Unpickler):
```
trivialfis: I have a more thorough helper script written here: https://github.com/dmlc/xgboost/pull/5281/files  The PR is still WIP, but expect to be finished soon.  Sorry for the caused troubles.

sh1ng:
@pseudotensor  There's xgboost-prev sub-module that point to https://github.com/h2oai/xgboost/tree/0.90-dmcl-update Both version must have the same dmcl version under the hood as xgboost uses singleton pattern to work with file system.  And when 2 versions of `libxgboost.so` get loaded in the same process call to the previous xgboost will actually invoke the new one(or vice  versa).

src/interface_py/h2o4gpu/util/xgboost_migration.py
sh1ng: Thank you for your help.
Probably I just copy-paste sklearn's stuff.

The goal of the script is to allow smoothly make conversion without saving model on disk and spawning a new process.

PR #837, opened 2020-04-10
Remove llvm and numba dependencies

sh1ng:


sh1ng:
It will fail probably as xgboost_prev has dependency on numba.

pseudotensor:
but numba import is in an unused function, so the import will never fail.

PR #838, opened 2020-04-10
Prepare support python 3.7, 3.8

sh1ng:
I prefer to merge it when tests for python 3.6 are passed as keeping it longer will slowdown CI because of docker cache misses.

* fixed swig wrapper
* LLVM 7.1 as we use quite old LLVM 6.0 that is not supported by python 3.7+
* pylint 

I'm going to add CI configuration for python 3.7 and 3.8 in a separate PR


Dockerfile-build
```diff
@@ -78,13 +72,23 @@ RUN \
     git checkout tags/v2.4.7-1 && \
     scl enable devtoolset-7 'make CUDA8_GENCODE="-gencode=arch=compute_35,code=sm_35 -gencode=arch=compute_50,code=sm_50 -gencode=arch=compute_60,code=sm_60 -gencode=arch=compute_61,code=sm_61" -j src.build';
 
+#
+# Library versions
+#
+ARG python_version=3.7
```
pseudotensor, line 78: is idea to switch to 3.7 entirely?  Or is this just testing things out?  DAI still supports 3.6 only.

Dockerfile-runtime
```diff
@@ -37,10 +37,12 @@ RUN bash -c 'if [ `arch` = "ppc64le" ]; then yum install -y git boost-devel cmak
 
 RUN mkdir -p /opt/h2oai/h2o4gpu
 
+ARG python_version=3.7
```
pseudotensor, line 40: here too, this is overall default now, but probably want it to be controllable which python version to build.

ci/Jenkinsfile-ppc64le-cuda10
```diff
@@ -30,6 +30,7 @@ def labelruntime = "ibm-power-gpu"
 def doingbenchmark = "0"
 def dobenchmark = "0"
 def doruntime = "1"
+def python = "3.6"
```
pseudotensor, line 33: Why 3.6 here?

sh1ng:
I was using default value `3.7` to tests correctness while CI script stays unchanged. Will remove it at all.

PR #845, opened 2020-05-26
Optional dependency on dask/dask-cuda/numba/llvm

sh1ng:


sh1ng:
@pseudotensor need to merge it ASAP

PR #854, opened 2020-07-24
fix device info

sh1ng:


src/gpu/utils.cu
```diff
@@ -104,14 +104,21 @@ int get_gpu_info_c(int verbose, int return_memory, int return_name,
...
-                  nvmlErrorString(rv));
-        return 1;
+      switch (rv) {
+        case NVML_SUCCESS:
+          gpu_percent_usage[i] = utilization.gpu;
+          break;
+        case NVML_ERROR_NOT_SUPPORTED:
+          log_warn(verbose, "Failed to get device %d utilization: %s", i,
+                   nvmlErrorString(rv));
+          gpu_percent_usage[i] = -1;
```
pseudotensor, line 114: Why -1?  Seems 0 is natural.

sh1ng: I'd prefer to use `nan` as it looks logical for unknown value, but return type is suppose to be int. Just wanted to indicate fact it's not 0

